PR #2, opened 2021-09-06
Initial commit of a Wasm parameter.

jetuk:
The new parameter takes a pre-compiled WASM file as input, and
runs it during a simulation using Wasmer. Parameter dependencies
are supported with values passed to the WASM function using
buffer.

Requires some additional error handling and some more extensive
tests.


jetuk:
This could be put behind a feature flag so `wasmer` is only an optional dependency.

PR #16, opened 2023-05-26
feat: Add initial implementation of PiecewiseStorage.

jetuk:
This is a working prototype of a PiecewiseStorage node that includes multiple internal stores. There is a basic unit test but limited functionality (e.g. no min volume, initial volume, or cost on the upper residual store).

### DIFF
--- a/src/metric.rs
+++ b/src/metric.rs
@@ -9,6 +9,23 @@ use crate::virtual_storage::VirtualStorageIndex;
 use crate::PywrError;
 
 #[derive(Clone, Debug, PartialEq)]
+pub struct VolumeBetweenControlCurves {
+    max_volume: Box<Metric>,
+    upper: Option<Box<Metric>>,
+    lower: Option<Box<Metric>>,
+}
+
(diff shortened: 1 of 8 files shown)


Batch21 (commented):
Had a play around with this and its looking good. Couple of things I noted:
 - It does look like its currently possible to set the control curve to a parameter. I tried this and got a custom nodes are not implemented error.
 - When saving outputs to H5, arrays are created for each piecewise store but not for the total storage.

jetuk:
Re the custom node error that is currently what happens if serde fails to parse your node definition. So, it could be a syntax error in defining the parameter. The custom node thing should perhaps be removed for now. Genrally, it might be preferable to have the node defined via a trait but that's tricky with the (de)serialisation.

Re outputs. This is related to the issue we spoke about before. Each node in the schema should expose a default attribute that is saved, in this case total volume. The current H5 recorder saves all the low level nodes in the model which is not intuitive for the user.

jetuk:
I've added a test with a proper control curve.

PR #23, opened 2023-06-26
[WIP] feat: Initial commit of 'outputs' section of schema.

jetuk:
Add optional "outputs" section to the model schema. Currently this supports defining a simple CSV or HDF5 output file.


jetuk:
Note quite happy with this:

- Possibly need an `OutputMetric` enum instead of trying to use the internal `Metric` one.
- Each schema node needs a method to return a given metric from a simple definition (or error if not supported).

PR #37, opened 2023-08-20
Cannot compile polars-core-0.30.0

s-simoncelli:
Hello,

When compiling the `polars-core` crate with `maturin`, the build fails. The problem does not occur after updating to `polars-0.32.0`. 

However the new version changed the signature of `to_ndarray` which is used in the `DataFrameParameter`. The function is now properly called with the `IndexOrder::default()` parameter (Fortran). This is the same behaviour in 0.30.0.


jetuk (approved):

jetuk:
Thanks for this update.

PR #40, opened 2023-09-01
Add division parameter

s-simoncelli:


src/parameters/division.rs
```diff
@@ -0,0 +1,53 @@
...
+    }
+    fn compute(
+        &self,
+        _timestep: &Timestep,
+        _scenario_index: &ScenarioIndex,
+        model: &Model,
+        state: &State,
+        _internal_state: &mut Option<Box<dyn Any + Send>>,
+    ) -> Result<f64, PywrError> {
+        // TODO handle scenarios
```
jetuk, line 43: I've just had a quick look at this.

I think there's some syntax errors here to do with the float constants. 

Pywr's current behaviour will return and unchecked division which will be a NaN. I think we should probably catch that here, like your implementation. There is a check for NaN in the parent code that calls `.compute()`, but it is a bit crude at the moment. 

Something like this:
```rust
        let denominator = self.denominator.get_value(model, state)?;

        if denominator == 0.0 {
            return Err(InvalidMetricValue(format!("Division by zero creates a NaN."));
        }

        let numerator = self.numerator.get_value(model, state)?;
        Ok(numerator / denominator)
```

src/schema/parameters/core.rs
```diff
@@ -93,6 +93,33 @@ impl MaxParameter {
...
+    pub fn add_to_model(
+        &self,
+        model: &mut crate::model::Model,
+        tables: &LoadedTableCollection,
+        data_path: Option<&Path>,
+    ) -> Result<ParameterIndex, PywrError> {
+        let n = self.numerator.load(model, tables, data_path)?;
+        let d = self.denominator.load(model, tables, data_path)?;
+
+        let p = crate::parameters::DivisionParameter::new(&self.meta.name, n, d);
```
jetuk, line 118: This should be `&self.meta.name()`

src/parameters/division.rs
```diff
@@ -0,0 +1,53 @@
...
+        let denominator = self.denominator.get_value(model, state)?;
+
+        if denominator == 0.0 {
+            return Err(InvalidMetricValue(format!("Division by zero creates a NaN in {}.", self.name)));
```
jetuk, line 47: This should be `self.name()`

jetuk:
`DivisionParameter` also does not exist in `pywr-schema` so we can't do the TryFromV1... trait yet.

jetuk:
I'll merge #45 and then merge or rebase this branch against main.

jetuk (approved):

PR #56, opened 2023-09-29
chore: Split crate into multiple crates.

jetuk:
- pywr-core: Contains the low-level model and interfaces with the LP solvers.
 - pywr-schema: Contains the Serde declared JSON schema.
 - pywr-python: Contains the Python extension module and package.
 - pywr-cli: Contains the Rust CLI.


pywr-cli/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "pywr-cli"
+version = "0.1.0"
```
Batch21, line 3: Should the version across the separate crates be kept in sync, like egui? Not sure I like the approach that much but might make the release process easier?

README.md
```diff
@@ -191,6 +191,14 @@ Feedback on porting models is very welcome, so please open an issue with any que
...
 <p align="right">(<a href="#readme-top">back to top</a>)</p>
 
+<!-- CRATES -->
+## Crates
+This repository contains the following crates:
+
+- `pywr-core`: A low-level Rust library for constructing network models. This crate interfaces with linear program solvers.
+- `pywr-schema`: A Rust library for validating Pywr JSON files against a schema, and then building a model from the schema using `pywr-core`.
+- `pywr-cli`: A command line interface for running Pywr models.
+- `pywr-python`: A Python package (and extension) for constructing and running Pywr models.
```
Batch21, line 201: Might also be worth updating the getting started instructions in the readme to say that you need to change directory to use the CLIs

Batch21 (approved):
Couple of minor suggestions but this looks good. Was able to run models using both CLIs

pywr-cli/Cargo.toml
jetuk: Good point. I don't think we'd necessarily need to keep them in-sync. We might just have to figure it out once we start releasing.

README.md
jetuk: Done

PR #58, opened 2023-10-09
feat: Schema Node improvements.

jetuk:
- Add NodeType as a simple enum.
- Add NodeBuilder for constructing new nodes.
- Impl Default for Node schema
- Remove CustomNode; don't want to support this going forward.


Batch21 (approved):
LGTM

PR #60, opened 2023-10-19
feat: Add support for multi-model simulation.

jetuk:
Add `PywrMultiModel` schema and corresponding `MultiNetworkModel` to pywr-core. This update refactors the schema of the regular model to use a new "network" key. The schema then provides a `PywrNetwork` containing only the nodes, edges, parameters, metric sets and outputs. This is then reused by the regular model and the multi-model.

A new parameter is added that can receive a parameter from another model. This is implemented via the transfers in the multi-model configuration. This is fairly generic because it uses the metric system to allowing passing of node or parameter data from one model to another.

TODO:

- [x] Write some documentation on the new structs.
- [x] Check naming convention for inter-model transfers.
- [x] Add a multi-model test that uses circular inter-model transfers.


pywr-cli/src/main.rs
```diff
@@ -173,48 +172,36 @@ fn run(path: &Path, solver: &Solver, data_path: Option<&Path>, output_path: Opti
     let data = std::fs::read_to_string(path).unwrap();
     let schema_v2: PywrModel = serde_json::from_str(data.as_str()).unwrap();
 
-    let (model, timestepper): (Model, Timestepper) = schema_v2.build_model(data_path, output_path).unwrap();
```
Batch21, line 176: looks like a cmd needs adding to run a multi-model. Something woth doing in this PR?

Batch21 (commented):
Latest changes look good and all the tests/models run succesfully for me. Made a few suggestions but happy if you want to get this merged before making additional changes.

pywr-core/src/models/multi.rs
```diff
@@ -0,0 +1,394 @@
...
+        timings.print_table();
+
+        Ok(())
+    }
+}
+
+/// Calculate inter-model parameters for the given scenario index.
+///
+///
+fn compute_inter_network_transfers(
```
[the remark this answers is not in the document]
jetuk: I think I'll add this as an issue. I have a plan for a longer example that would require passing index (integer) values between models as well.

pywr-cli/src/main.rs
```diff
@@ -116,6 +132,15 @@ fn main() -> Result<()> {
...
             } => run(model, solver, data_path.as_deref(), output_path.as_deref(), *debug),
+            Commands::RunMulti {
+                model,
+                solver,
+                data_path,
+                output_path,
+                parallel,
+                threads,
+                debug,
+            } => run(model, solver, data_path.as_deref(), output_path.as_deref(), *debug),
```
Batch21, line 143: should be `run_multi`?

Batch21 (approved):

PR #71, opened 2023-11-21
feat: Initial implementation of virtual storage costs.

jetuk:
Virtual storage nodes now add a reference to their linked flow nodes. Those nodes now include an function that aggregates over the local cost and the costs of any linked virtual storage nodes.

There's currently no method to change the default aggregation function from "max" via the schema.


Batch21 (approved):
Looks good. A couple of suggestion for future updates:
 - This is obviously going to need some good documentation especially as its a pretty significant change from v1.
 - The ability to specify separate agg funcs to apply to the virtual costs and to the aggregated virtual cost and the local cost could be useful.

PR #73, opened 2023-11-28
feat: Refactor the Python API.

jetuk:
Remove the functions and replace with two wrapper classes for Schema and Model. These have basic functionality to deserialise, load and run a model.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,13 +118,15 @@ git submodule update
 ```
 
 Rust is required for installation of the Python extension. To create a Python development installation
-requires first compiling the Rust library and then installing the Python package in editable model. The
-following example uses [Poetry](https://python-poetry.org/) to manage the Python environment, compile
-the Pywr extension and run the Pywr Python CLI.
+requires first compiling the Rust library and then the Python extension. The following example uses
+a virtual environment to install the Python dependencies, compile the Pywr extension and run the Pywr Python CLI.
+
 
 ```bash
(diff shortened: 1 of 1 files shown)


jetuk:
> Had a quick look through this and I like the approach. A few observations:
> 
>     * Python cli is now broken though I suspect that fixing it is already on your TODO list..

I hadn't actually noticed this. Do we need a Rust CLI and a Python CLI? If this is staying it needs adding to the Python tests. The pipeline also needs re-enabling.

>     * It would be useful to add some dev dependencies to pyproject toml - pyarrow, h5py, etc..

Noted!

>     * The piecewise-link1 test failed for me. It could not find the `mrf1/step-00` component in the output h5 file.

Yes, this is why this is WIP. The test wanted to check the routing through the piecewise link. For this you need to save the sub-node timeseries. This needs support in the `OutputMetric` enum.

jetuk:
This is waiting on #85 for refactoring.

README.md
```diff
@@ -117,21 +117,16 @@ git submodule init
...
-pip install -e .
-```
-
-Alternatively use the `develop.sh` script to run the above two commands.
-
-Once this is complete the following will run a simple command line application from Python.
-
-```bash
-python -m pywr
+poetry install # install dependencies
```
Batch21, line 126: This will no longer work with the poetry sections taken out of pyproject.toml?

PR #76, opened 2023-11-30
chore: Update the Rust CLI binary to "pywr".

jetuk:
It was previously called `pywr-cli` after the crate name. This makes it just `pywr`. Also updated the installation instructions.


jetuk:
This is failing because Windows does not like the binary `pywr` having the same name as the Python extension. I have no idea why.

PR #77, opened 2023-12-11
feat: Improve the metric set implementation.

jetuk:
This now includes nested aggregation, hooking up the metric sets to the model and use by the CSV recorder.

There's still some WIP but a general review would be good at this point.

### DIFF
--- a/pywr-core/src/model.rs
+++ b/pywr-core/src/model.rs
@@ -5,8 +5,8 @@ use crate::edge::{EdgeIndex, EdgeVec};
 use crate::metric::Metric;
 use crate::node::{ConstraintValue, Node, NodeVec, StorageInitialVolume};
 use crate::parameters::{MultiValueParameterIndex, ParameterType};
-use crate::recorders::{MetricSet, MetricSetIndex};
-use crate::scenario::{ScenarioGroupCollection, ScenarioIndex};
+use crate::recorders::{MetricSet, MetricSetIndex, MetricSetState};
+use crate::scenario::{ScenarioGroup, ScenarioGroupCollection, ScenarioIndex};
 use crate::solvers::{MultiStateSolver, Solver, SolverFeatures, SolverSettings, SolverTimings};
 use crate::state::{ParameterStates, State};
 use crate::timestep::{Timestep, Timestepper};
@@ -169,6 +169,7 @@ impl Model {
             Vec<ScenarioIndex>,
             Vec<State>,
             Vec<ParameterStates>,
+            Vec<Vec<MetricSetState>>,
             Vec<Option<Box<dyn Any>>>,
         ),
         PywrError,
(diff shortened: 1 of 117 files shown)


pywr-core/src/recorders/aggregator.rs
```diff
@@ -42,14 +42,10 @@ impl AggregationPeriod {
         while current_date < end_date {
             // This should be safe to unwrap as it will always create a valid date unless
             // we are at the limit of dates that are representable.
```
Batch21, line 44: by the time crate?

pywr-core/src/recorders/aggregator.rs
```diff
@@ -176,44 +204,127 @@ impl PeriodValue {
...
+            self.agg.process_value(&mut state.state, agg_value)
+        } else {
+            None
+        }
+    }
+
+    /// Compute the final aggregation value from the current state.
+    ///
+    /// This will also compute the final aggregation value from the child aggregators if any exists.
+    /// This includes aggregation calculations over partial or unfinished periods.
```
Batch21, line 290: Should this be a user setting?

pywr-schema/src/metric_sets/mod.rs
```diff
@@ -28,11 +29,82 @@ impl OutputMetric {
...
+            value.child.map(|a| (*a).into()),
+        )
+    }
+}
+
+/// A set of metrics that can be output from a model run.
+///
+/// A metric set can optionally have an aggregator, which will apply an aggregation function
+/// over metrics set. If the aggregator has a defined frequency then the aggregation will result
+/// in multiple values (i.e. per each period implied by the frequency).
```
Batch21, line 102: I think it would be good to have a doc text model here that shows how the nest aggregation works

Batch21 (commented):
Given this a good read through. I like the aggregation approach and it looks like it works correctly.

pywr-core/src/model.rs
```diff
@@ -311,8 +316,13 @@ impl Model {
...
 
         // Setup the model and create the initial state
-        let (scenario_indices, mut states, mut parameter_internal_states, mut recorder_internal_states) =
-            self.setup(&timesteps)?;
+        let (
+            scenario_indices,
+            mut states,
+            mut parameter_internal_states,
+            mut metric_set_internal_states,
+            mut recorder_internal_states,
```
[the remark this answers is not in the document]
jetuk: Yes, this is something that needs doing. I was planning to tidy it up after this PR

pywr-core/src/recorders/aggregator.rs
jetuk: Actually, this comment is old and should be moved to the `start_of_next_period` function.

pywr-core/src/recorders/aggregator.rs
jetuk: I'll add this as a new issue for the sake of getting this PR done.

pywr-schema/src/metric_sets/mod.rs
jetuk: I wonder if this would be better as an extended section in the proposed book? I am tempted to leave it for now, but then also need to ensure we write that documentation.

jetuk:
I've fixed the out-of-date comment. This could be merged now, or we can address wider documentation issue first (at the expense of delaying the merge).

Batch21 (approved):

PR #84, opened 2024-01-15
feat: Implement Clone for NetworkState.

jetuk:
This requires implementing Clone for parameters' internal state objects. For this a new super trait, ParameterState, is added that requires Clone to be implemented. This is achieved via the dyn_clone crate.

The Wasmer Store is not Clone which would block this function. As the WASM parameter was a proof-of-concept this has been removed for the time being in order to not block this functionality.


Batch21 (approved):
LGTM

PR #85, opened 2024-01-22
feat: Initial implementation of NodeAttribute

jetuk:
This is a refactor of NodeReference to specify a schema facing attribute (NodeAttribute). This idea is to remove the need for the user of the schema to understand the internals of complex nodes. Instead they will specify a node by its name in the schema and an associated NodeAttribute.

NodeAttribute is intended as a global enum of all possible values. However, each node may only support a subset of these attributes. I.e. some of the attributes do not make sense for certain node types.

This has largely worked. However, there is an ugly requirement in the control curves to specify the NodeReference struct where it was previously just a string.

### DIFF
--- a/pywr-core/benches/random_models.rs
+++ b/pywr-core/benches/random_models.rs
@@ -20,8 +20,6 @@ use pywr_core::test_utils::make_random_model;
 use rand::SeedableRng;
 use rand_chacha::ChaCha8Rng;
 
-
-
 fn random_benchmark(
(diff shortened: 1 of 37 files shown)


Batch21 (changes requested):
Couple of comments but otherwise looks good.

I don't think the control curve change is too ugly. Little bit more explicit and probably means that a user could use control curves with absolute values if they wanted?

Batch21 (approved):

PR #87, opened 2024-01-25
chore: use an enum for the initial_volume attribute of the schema StorageNode

Batch21:
Previously the node had two attributes, initial_volume and initial_volume_pc. These are now variants of the enum.


Batch21:
@jetuk there are a couple of implementation details here I'm not sure about:
- I've re-implemented the enum in the schema to avoid adding serde as a dependency to core. However, I've had to import the core version so that the schema version can be converted to it. An alternative would be to import the schema version into core and delete the version there, though this would involve adding the schema as a dependency of core.
- Should the default initial volume be full or empty.

pywr-schema/src/nodes/core.rs
```diff
@@ -313,15 +313,37 @@ impl TryFrom<OutputNodeV1> for OutputNode {
     }
 }
 
-#[derive(serde::Deserialize, serde::Serialize, Clone, Default)]
+#[derive(serde::Deserialize, serde::Serialize, Clone, PartialEq, Copy, Debug)]
+pub enum StorageInitialVolume {
+    #[serde(alias = "absolute")]
```
jetuk, line 318: I don't think we've generally renamed or aliased anything in the new schema. E.g. node types are now required to be capitalised. I wonder if, for consistency, we remove this and use the serde default names. Thoughts?

jetuk:
This LGTM. The conflicts need addressing, and it will need implementing in #88 or here depending on which gets merged first.

pywr-schema/src/nodes/core.rs
Batch21: I agree, its better to be consistent

pywr-schema/src/nodes/annual_virtual_storage.rs
```diff
@@ -167,15 +159,22 @@ impl TryFrom<AnnualVirtualStorageNodeV1> for AnnualVirtualStorageNode {
             .map(|v| v.try_into_v2_parameter(Some(&meta.name), &mut unnamed_count))
             .transpose()?;
 
+        let initial_volume = if let Some(v) = v1.initial_volume {
+            StorageInitialVolume::Absolute(v)
+        } else if let Some(v) = v1.initial_volume_pc {
+            StorageInitialVolume::Proportional(v)
+        } else {
+            StorageInitialVolume::default()
```
jetuk, line 167: Should this return a `Err(ConversionError::MissingAttribute)`? Or does Pywr v1 not require this attribute?

jetuk:
The Python test models need updating to the new enum format.

jetuk (approved):

PR #88, opened 2024-01-26
feat: Implement RollingVirtualStorage node.

jetuk:
An initial implementation of RollingVirtualStorage node.

- Adds support for rolling history in core virtual storage.
- Adds schema for RollingVirtualStorage.
- Implements conversion from v1 format.
- Basic test model.


pywr-schema/src/nodes/rolling_virtual_storage.rs
```diff
@@ -0,0 +1,247 @@
...
+        };
+
+        let node_idxs = self
+            .nodes
+            .iter()
+            .map(|name| network.get_node_index_by_name(name.as_str(), None))
+            .collect::<Result<Vec<_>, _>>()?;
+
+        // The rolling licence never resets
+        let reset = VirtualStorageReset::Never;
```
Batch21, line 113: Is there a possibility that a user might want it reset? e.g. a rolling licences that only operates over the summer

Batch21: Though maybe this would be a separate `SeasonalRollingVirtualStorage` scheme type as you'd need away to turn it off

Batch21 (changes requested):
implementation looks good. A few comments/suggestions:

pywr-core/src/timestep.rs
```diff
@@ -82,6 +82,14 @@ pub struct TimeDomain {
 }
 
 impl TimeDomain {
+    /// Return the duration of each time-step.
+    pub fn step_duration(&self) -> Duration {
+        // This relies on the assumption that all time-steps are the same length.
+        // Ideally, this invariant would be refactored to have the duration stored here in `TimeDomain`,
+        // rather than in `Timestep`.
```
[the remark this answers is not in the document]
jetuk: https://github.com/pywr/pywr-next/issues/89

pywr-schema/src/nodes/rolling_virtual_storage.rs
jetuk: I think, for now, this mirrors old behaviour. We can add to this basic functionality in the future.

Batch21 (approved):

Issue #89, opened 2024-01-28
Time domain improvements


jetuk:
The current time domain has some limitations which could be addressed:

- The time-step is an integer number of days. It needs to to handle calendar based time-steps to match Pywr v1 support.
- The above complicates some calculations (e.g. number of days in rolling windows) when using calendar / varying length time-steps.
- The current implementation has a fixed time-step, but this is only accessible by looking at individual `Timestep` structs. I.e. the invariant that it is a fixed time-step could be better represented.

PR #99, opened 2024-02-07
feat: add timeseries section to model schema

Batch21:


Batch21:
@jetuk there is quite a bit left to do here but it would be good to get your opinion on the approach.

I have a few queries:

* Should `Timeseries` be a variant of `MetricFloatValue`? Originally I made it a variant of `MetricFloatReference` but as the load method of this enum is called in the model build method when setting up the inter-network transfers, it required creating an empty timeseries collection, which was a bit ugly. 
* Should the timeseries inputs be able to handle index timeseries? The `load_column` and `load_df` methods could probably take a generic type arg for loading either floats or ints.
* Should timeseries and tables be combined into a single inputs section?

pywr-schema/src/timeseries/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+struct ResamplingArgs {}
+
+#[derive(serde::Deserialize, serde::Serialize, Debug, Clone)]
+pub struct Timeseries {
+    #[serde(flatten)]
+    names: ParameterMeta,
+    provider: TimeseriesProvider,
+    url: PathBuf,
```
jetuk, line 33: I think we discussed this might be defined in each provider as required. There might be providers that do not need a URL.

pywr-schema/src/timeseries/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+    Pandas,
+    Polars(PolarsDataset),
+}
+
+struct ResamplingArgs {}
+
+#[derive(serde::Deserialize, serde::Serialize, Debug, Clone)]
+pub struct Timeseries {
+    #[serde(flatten)]
+    names: ParameterMeta,
```
jetuk, line 31: Should be `meta`?

pywr-schema/src/timeseries/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+impl Timeseries {
+    pub fn load(&self, domain: &ModelDomain, data_path: Option<&Path>) -> Result<DataFrame, SchemaError> {
+        match &self.provider {
+            TimeseriesProvider::Polars(dataset) => dataset.load(self.url.as_path(), data_path),
+            TimeseriesProvider::Pandas => todo!(),
+        }
+    }
+}
+
+fn align_and_resample(df: &DataFrame, domain: &ModelDomain) -> Result<DataFrame, SchemaError> {
```
jetuk, line 45: This is what I imagined this might look like. Perhaps it returns an `Array2` instead of `DataFrame`?

I think we might want this in a separate module with a bunch of separate unit tests. 

Also, I guess there's some convention about the format / structure of a time column in `df`?

pywr-schema/src/timeseries/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+    // 5. slice df to align with domain start and end dates
+
+    todo!()
+}
+
+pub struct LoadedTimeseriesCollection {
+    timeseries: HashMap<String, DataFrame>,
+}
+
+impl LoadedTimeseriesCollection {
```
jetuk, line 59: This looks like good to me. So, basically the timeseries get copied into an Array parameter. This will be inefficient if there are multiple references to the same data. Each will create a new array in the model.  That's probably good enough for now. 

I think the only way to improve upon that would be replicate the idea of a timeseries in core. Perhaps this could be something like the array parameters. If they implemented `MultiValueParameter` it might allow you to create an `Array2Parameter` and then refer to specific columns?

pywr-schema/src/timeseries/polars_dataset.rs
```diff
@@ -0,0 +1,56 @@
...
+        let fp = if url.is_absolute() {
+            url.to_path_buf()
+        } else if let Some(data_path) = data_path {
+            data_path.join(url)
+        } else {
+            url.to_path_buf()
+        };
+
+        let df = match fp.extension() {
+            Some(ext) => match ext.to_str() {
```
jetuk, line 24: Does this need to be lower cased?

pywr-schema/src/timeseries/polars_dataset.rs
```diff
@@ -0,0 +1,56 @@
...
+                Some("csv") => CsvReader::from_path(fp)?
+                    .infer_schema(None)
+                    .with_try_parse_dates(true)
+                    .has_header(true)
+                    .finish()?,
+                Some("parquet") => {
+                    todo!()
+                }
+                Some(other_ext) => {
+                    return Err(SchemaError::TimeseriesUnsupportedFileFormat {
```
jetuk, line 34: You could create a `TimeseriesError` enum for this module? I've thought about that more generally.

pywr-schema/src/nodes/annual_virtual_storage.rs
```diff
@@ -53,24 +54,49 @@ impl AnnualVirtualStorageNode {
         tables: &LoadedTableCollection,
         data_path: Option<&Path>,
         inter_network_transfers: &[PywrMultiNetworkTransfer],
+        timeseries: &LoadedTimeseriesCollection,
     ) -> Result<(), SchemaError> {
         let cost = match &self.cost {
             Some(v) => v
-                .load(network, schema, domain, tables, data_path, inter_network_transfers)?
+                .load(
```
jetuk, line 61: These arguments are bit silly now. Not for this PR, but it would be a good idea to think about combining these into a struct at some point.

jetuk (commented):
I've had a quick read through. I've left a couple of comments. I think this is what I was thinking it would look like.

Batch21:
@jetuk I've had a go at some resampling using polars. It still needs some work but one change that might simplify the code a bit would be to use chrono across the project instead of time-rs. This would also open up the possibility of using polars duration [strings](https://docs.rs/polars/latest/polars/prelude/struct.Duration.html#method.parse) to specify model timestep frequency and the polars `date_range` function to generate the timesteps. What do you think?

jetuk:
I think the polars integration and the duration strings are a good argument for switching to chrono. How should we do that? Should we do a separate PR to swap, and the rebase this against it?

Batch21:
@jetuk could you take another look at this. I responded to most of your comments thought there are still some issues to fix:

- Resampling to or from timesteps with non-uniform durations does not currently work:
	* If the input data has a non-uniform datetime column the code will currently hit a `todo!()`. One issue with implementing something here is that polars does not have a direct way to infer frequency from a datetime column. I can't find any other rust crates with this functionality but looking at the Pandas code for it, it might not be too hard to implement it for a few frequencies (maybe we only need monthly initially?).
	* For the model timesteps, currently the `TimeDomain` derives step duration using only the first timestep so will not account for non-uniform timestep durations. This needs fixing alongside the last part of #89, which I think you said you wanted to have a look at? 
- Its probably unnecessary to use nanosecond precision when comparing the durations. Seconds is probably fine?
- I've fixed an issue where a `ParameterNameAlreadyExist` error occurred if a dataframe or dataframe column was referenced more than once. However there is still the issue that a user could define a parameter with a name that matches with one that would be created for a timeseries input. In this case, it would now use the values from the user-defined parameter if this was loaded first and not load the timeseries data. I wonder if parameters should have some kind of namespacing so that we can easily separate user-defined parameters from those created internally. 

At least the first of these is probably worth doing in another PR?

jetuk:
> @jetuk could you take another look at this. I responded to most of your comments thought there are still some issues to fix:
> 
>     * Resampling to or from timesteps with non-uniform durations does not currently work:
>       
>       * If the input data has a non-uniform datetime column the code will currently hit a `todo!()`. One issue with implementing something here is that polars does not have a direct way to infer frequency from a datetime column. I can't find any other rust crates with this functionality but looking at the Pandas code for it, it might not be too hard to implement it for a few frequencies (maybe we only need monthly initially?).

That's probably OK. There are other TODOs and unfinished items at the moment. We can get this framework in and work on it.

>       * For the model timesteps, currently the `TimeDomain` derives step duration using only the first timestep so will not account for non-uniform timestep durations. This needs fixing alongside the last part of [Time domain improvements #89](https://github.com/pywr/pywr-next/issues/89), which I think you said you wanted to have a look at?

Yes, I had some ideas, but nothing implemented. Can we merge this with a limitation? That method could do a check that all time-steps have the same duration and panic if not?

> 
>     * Its probably unnecessary to use nanosecond precision when comparing the durations. Seconds is probably fine?
> 

Yes!

>     * I've fixed an issue where a `ParameterNameAlreadyExist` error occurred if a dataframe or dataframe column was referenced more than once. However there is still the issue that a user could define a parameter with a name that matches with one that would be created for a timeseries input. In this case, it would now use the values from the user-defined parameter if this was loaded first and not load the timeseries data. I wonder if parameters should have some kind of namespacing so that we can easily separate user-defined parameters from those created internally.

This could mirror the "sub-names" for the nodes? I hadn't anticipated a schema parameter creating many core parameters. 
 

> At least the first of these is probably worth doing in another PR?

Probably all of them can be addressed with additional updates?

Batch21:
@jetuk this is ready for another look. All your original comments should be addressed in some fashion.

One update that's probably necessary is adding the `parameters_mut` method to all the nodes. However, I wanted to check first if you though there was a better approach?

jetuk:
I think there are other approaches, but they will delay this PR further. 

Two ideas:

1. Remove the manually writing of `parameters` and `parameters_mut` using a derive macro (see v1 schema). 
2. Move to a builder pattern for mutating nodes (I'm not sure about this).
3. Use a visitor pattern instead. Here there would be a `visit` or `visit_mut` method which you would give a closure that's called with each field or parameter. I am more used to (1), but thought about this after reading some of the tracing stuff. 

So, probably (1) would just remove this boiler plate. I guess it might be more efficient to do (1) in a separate PR and then update this one? Assuming we can do that in the short-term to not delay this further.

Batch21:
I forgot that the proc macros you created in the v1 schema did this. Looking at them, we should be able to copy them across without too many modifications. I can have a go in a separate PR this evening.

pywr-core/src/timestep.rs
```diff
@@ -58,6 +58,10 @@ impl PywrDuration {
     pub fn fractional_days(&self) -> f64 {
         self.0.num_seconds() as f64 / SECS_IN_DAY as f64
     }
+
```
jetuk, line 61: Add docstring. Would be good to make the one above a proper docstring as well.

pywr-schema/src/model.rs
```diff
@@ -412,8 +445,9 @@ impl PywrModel {
             Timestepper::default()
         });
 
-        let nodes = v1
+        let nodes_and_ts: Vec<NodeAndTimeseries> = v1
```
jetuk, line 448: It might be worthwhile adding some comments here about this process.

pywr-schema/src/model.rs
```diff
@@ -424,22 +458,29 @@ impl PywrModel {
...
-                    })
-                    .collect::<Vec<_>>(),
-            )
+            let (parameters, param_ts_data) = convert_parameter_v1_to_v2(v1_parameters, &mut unnamed_count, &mut errors);
+            (Some(parameters), Some(param_ts_data))
+        } else {
+            (None, None)
+        };
+
+        ts_data.extend(param_ts_data.into_iter().flatten());
```
jetuk, line 479: I think you can move this up into the above `if let Some(v1_parameters) ...` block to avoid wrapping in an `Option`?

pywr-schema/src/timeseries/align_and_resample.rs
```diff
@@ -0,0 +1,245 @@
...
+    let durations = df
+        .clone()
+        .lazy()
+        .select([col(time_col).diff(1, NullBehavior::Drop).unique().alias("duration")])
+        .collect()?;
+    let durations = durations.column("duration")?.duration()?.deref();
+
+    if durations.len() > 1 {
+        // Non-uniform timestep are not yet supported
+        todo!();
```
jetuk, line 35: Put the comment in the `todo!()` macro.

pywr-schema/src/timeseries/align_and_resample.rs
```diff
@@ -0,0 +1,245 @@
...
+    let timeseries_duration = match durations.get(0) {
+        Some(duration) => duration,
+        None => return Err(TimeseriesError::TimeseriesDurationNotFound(name.to_string())),
+    };
+
+    let model_duration = domain
+        .time()
+        .step_duration()
+        .whole_nanoseconds()
+        .expect("Nano seconds could not be extracted from model step duration");
```
jetuk, line 47: Should this be a `TimesreiesError` variant?

pywr-schema/src/timeseries/align_and_resample.rs
```diff
@@ -0,0 +1,245 @@
...
+                        start_by: StartBy::DataPoint,
+                        ..Default::default()
+                    },
+                )
+                .agg([col("*").exclude([time_col]).mean()])
+                .collect()?
+        }
+        Ordering::Less => {
+            // Upsample
+            // TODO: this does not extend the dataframe beyond its original end date. Should it do when using a forward fill strategy?
```
jetuk, line 70: Yes, I think so.

pywr-schema/src/timeseries/align_and_resample.rs
```diff
@@ -0,0 +1,245 @@
...
+            df.clone()
+                .upsample::<[String; 0]>([], "time", Duration::new(model_duration), Duration::new(0))?
+                .fill_null(FillNullStrategy::Forward(None))?
+        }
+        Ordering::Equal => df,
+    };
+
+    let df = slice_end(df, time_col, domain)?;
+
+    // TODO check df length equals number of model timesteps
```
jetuk, line 81: This seems quite an important check to implement soon?

pywr-schema/src/timeseries/align_and_resample.rs
```diff
@@ -0,0 +1,245 @@
...
+            "time",
+            NaiveDate::from_ymd_opt(2021, 1, 1).unwrap().into(),
+            NaiveDate::from_ymd_opt(2021, 1, 31).unwrap().into(),
+            Duration::parse("1d"),
+            ClosedWindow::Both,
+            TimeUnit::Milliseconds,
+            None,
+        )
+        .unwrap();
+        //let values: Vec<f64> = vec![1.0; 31];
```
jetuk, line 132: Remove

pywr-schema/src/timeseries/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+    #[serde(flatten)]
+    meta: ParameterMeta,
+    provider: TimeseriesProvider,
+}
+
+impl Timeseries {
+    pub fn load(&self, domain: &ModelDomain, data_path: Option<&Path>) -> Result<DataFrame, TimeseriesError> {
+        match &self.provider {
+            TimeseriesProvider::Polars(dataset) => dataset.load(self.meta.name.as_str(), data_path, domain),
+            TimeseriesProvider::Pandas => todo!(),
```
jetuk, line 58: We should open an issue for this.

pywr-schema/src/timeseries/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+    pub fn from_schema(
+        timeseries_defs: Option<&[Timeseries]>,
+        domain: &ModelDomain,
+        data_path: Option<&Path>,
+    ) -> Result<Self, TimeseriesError> {
+        let mut timeseries = HashMap::new();
+        if let Some(timeseries_defs) = timeseries_defs {
+            for ts in timeseries_defs {
+                let df = ts.load(domain, data_path)?;
+                // TODO error if key already exists
```
jetuk, line 82: Let's implement this to avoid weird errors.

pywr-schema/src/timeseries/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+                    },
+                );
+            }
+            TimeseriesV1Source::Url(url) => {
+                let name = data.name.as_ref().unwrap().clone();
+                if ts.contains_key(&name) {
+                    continue;
+                }
+
+                // TODO: time_col should use the index_col from the v1 data?
```
jetuk, line 178: Yes

pywr-schema/src/error.rs
```diff
@@ -1,5 +1,7 @@
 use crate::data_tables::TableError;
 use crate::nodes::NodeAttribute;
+use crate::timeseries::TimeseriesError;
+use polars::error::PolarsError;
```
jetuk, line 4: Unused import

pywr-schema/src/model.rs
```diff
@@ -4,12 +4,21 @@ use super::parameters::Parameter;
 use crate::data_tables::{DataTable, LoadedTableCollection};
 use crate::error::{ConversionError, SchemaError};
 use crate::metric_sets::MetricSet;
+use crate::nodes::NodeAndTimeseries;
 use crate::outputs::Output;
-use crate::parameters::{MetricFloatReference, TryIntoV2Parameter};
+use crate::parameters::{
```
jetuk, line 9: Some unused imports to resolve here.

pywr-schema/src/parameters/control_curves.rs
```diff
@@ -5,6 +5,7 @@ use crate::nodes::NodeAttribute;
 use crate::parameters::{
     DynamicFloatValue, IntoV2Parameter, NodeReference, ParameterMeta, TryFromV1Parameter, TryIntoV2Parameter,
 };
+use crate::timeseries::{self, LoadedTimeseriesCollection};
```
jetuk, line 8: Unused import `self`

pywr-schema/src/parameters/offset.rs
```diff
@@ -1,5 +1,6 @@
 use crate::data_tables::LoadedTableCollection;
-use crate::parameters::{ConstantValue, DynamicFloatValue, DynamicFloatValueType, ParameterMeta};
+use crate::parameters::{ConstantValue, DynamicFloatValue, DynamicFloatValueType, ParameterMeta, VariableSettings};
```
jetuk, line 2: Unused import `VariableSettings`

pywr-schema/src/timeseries/align_and_resample.rs
```diff
@@ -0,0 +1,245 @@
+use chrono::TimeDelta;
```
jetuk, line 1: Unused import

pywr-schema/src/timeseries/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+
+use ndarray::Array2;
+use polars::error::PolarsError;
+use polars::prelude::DataType::Float64;
+use polars::prelude::{DataFrame, Float64Type, IndexOrder};
+use pywr_core::models::ModelDomain;
+use pywr_core::parameters::{Array1Parameter, Array2Parameter, ParameterIndex};
+use pywr_core::PywrError;
+use pywr_v1_schema::tables::TableVec;
+use std::sync::Arc;
```
jetuk, line 12: Unused import

jetuk (changes requested):
This looks good. I've left a few, but minor, comments to improve docstrings and tidy things up etc. 

I think this might need a `cargo fmt` too.

pywr-schema/src/timeseries/align_and_resample.rs
Batch21: I might do this one in a new PR if that's ok as its going to require a bit of testing

Batch21:
@jetuk thanks for the review. I think I've addressed all your comments apart from the one I reply to above.

pywr-schema/src/timeseries/mod.rs
jetuk: What about this one?

Batch21: oops I started it but didn't finish. Should be done now

jetuk:
Should we remove `DataframeParameter` in this PR? Or mark it deprecated in some fashion?

Batch21:
> Should we remove `DataframeParameter` in this PR? Or mark it deprecated in some fashion?

yeah, I did think of doing this earlier on. At this stage of development it could probably just be removed?

Batch21:
> @Batch21 Are you finished with this now?

Yes, though I did spot another issue when removing the df parameter. The polars timeseries backend does not load gzip files, which will be needed when the test that loads the `simple-wasm` test model is enabled. I can add an issue for this.

jetuk (approved):

Issue #53, opened 2023-09-27
Should the variable API mutate Model?


jetuk:
The `VariableParameter` trait currently contains this method signature. 

```rust
    fn set_variables(&mut self, values: &[f64]) -> Result<(), PywrError>;
```

This requires self be mutable. The idea being that updating a parameter's values from an external source (e.g. optimisation algorithm) will modify the parameter data itself. During simulation a model (and its parameters) are immutable, with separate state created and stored for the duration of the simulation. This separation of model and state is neat and useful for various reason. 

Should the `VariableParameter` API also mutate some separate state rather than the model itself. I.e. the model would remain immutable and require another accompanying state to hold the current value(s) of any "variable" data for the parameters. This could be the existing internal state available to parameters. 

For example, the above method could be written as follows.

```rust
    fn set_variables(&self, values: &[f64], internal_state: &mut Option<Box<dyn Any + Send>>) -> Result<(), PywrError>;
```

Self is immutable, and would require any implementation of this trait to store the updated values in the internal state.

jetuk:
I was looking at this again today. One issue is that there is an internal state value for each scenario. It does not feel correct to update this variable data for each scenario. However, maybe this does open the possibility of an RDM-like like process which sets a different value for a variable in each scenario based on the upper and lower bounds.

There is definitely benefit to having the model be immutable. It will save a lot of memory when running multi-process workers that do not have to clone the entire model (i.e. all the boundary condition data). 

The main issue here is the added complexity of storing the modified values in the internal state objects. Take, for example, The RBF profile implementation. At the moment this pre-calculates the daily profile from the provided points. When updating the point values via the variable API both the new point values and the updated profile need to be stored in the internal state. This adds quite a bit of complexity to the internal implementation.

PR #100, opened 2024-02-09
Immutable model var api

jetuk:


jetuk:
I think this is ready for a review. The general idea is to solve #53 . The purpose being to make the reference to the model instance immutable when applying a value through the variable API. I.e those changes are made to the state. 

This should have advantages when multi-threading and using the variable API (e.g. MOEA). The separation of data is quite nice. However, it does complicate the parameter definitions in core. 

It would be nice to eventually expose this API via the Python parameters, but that is for another time.

Batch21 (changes requested):
A few queries but this looks good

jetuk:
I've addressed the comments, thanks.

Batch21 (approved):

PR #105, opened 2024-02-12
Added hydropower parameter in core and schema

s-simoncelli:


jetuk:
Thanks for this. A quick comment to say that for outputting hydropower I think this should become a derived metric (like deficit). I think this would be different to the hydropower recorders in Pywr v1. Eventually, I would also like to add a `Turbine` node that has power has an attribute for recorder/output.

s-simoncelli:
The parameter and node implementation is moved to https://github.com/pywr/pywr-next/pull/105

PR #115, opened 2024-02-15
WIP: feat: switch from time-rs to chrono and implement non-daily timesteps

Batch21:


Batch21:
@jetuk would be good to get some feedback on this. One potential change I've just noticed is that the virtual storage state will need updating to store timestep length, otherwise there will be no way to know the amount of flow to recover for each timestep.

jetuk:
> One potential change I've just noticed is that the virtual storage state will need updating to store timestep length, otherwise there will be no way to know the amount of flow to recover for each timestep.

I'm not sure I understand? The current implementation uses `Timestep.days()` the same as the regular storage state?

jetuk (commented):
I've had a quick read through and left some comments.

Batch21:
> > One potential change I've just noticed is that the virtual storage state will need updating to store timestep length, otherwise there will be no way to know the amount of flow to recover for each timestep.
> 
> I'm not sure I understand? The current implementation uses `Timestep.days()` the same as the regular storage state?

In `recover_virtual_storage_last_historical_flow` the current timestep is used but its length might be different to the timestep when the flow added to the storage state?

jetuk:
I understand. It seems like the history would need a method to pop off mean
flow over a period equal to the current time-step duration. And yes,
therefore it would need to keep a record of the durations associated with
each history entry.

I think generally this will be ok as it is because time-steps tend to be
similar in duration if bit identical. Therefore, note it as an issue and
we'll fix it in another PR.

pywr-core/src/parameters/discount_factor.rs
```diff
@@ -5,6 +5,7 @@ use crate::scenario::ScenarioIndex;
 use crate::state::{ParameterState, State};
 use crate::timestep::Timestep;
 use crate::PywrError;
+use chrono::Datelike;
```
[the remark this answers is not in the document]
Batch21: Yeah, its the trait that provides methods such as `day` and `year` on `NaiveDate` and `NaiveDateTime` so needs to be imported in quite a few places.

pywr-core/src/timestep.rs
```diff
@@ -46,34 +42,89 @@ impl Add<Duration> for Timestep {
...
+    Days(i64),
+    Frequency(String),
+}
+
 #[derive(Debug)]
 pub struct Timestepper {
-    start: Date,
-    end: Date,
-    timestep: Duration,
+    start: NaiveDateTime,
```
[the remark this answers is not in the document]
Batch21: Yes, to make sub-daily timesteps possible. If we didn't want to allow these at this point we could revert to a `NaiveDate`

pywr-schema/src/model.rs
```diff
@@ -48,20 +49,40 @@ impl From<pywr_v1_schema::model::Timestep> for Timestep {
     }
 }
 
+#[derive(serde::Deserialize, serde::Serialize, Clone, Copy)]
+pub struct DateTimeComponents {
```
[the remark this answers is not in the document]
Batch21: yes, at a minimum I think the string has to have hour and minute components

pywr-core/src/parameters/profiles/monthly.rs
```diff
@@ -30,10 +30,20 @@ impl MonthlyProfileParameter {
     }
 }
 
+fn days_in_year_month(datetime: &NaiveDateTime) -> u32 {
+    match datetime.month() {
```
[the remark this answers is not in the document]
Batch21: > Use the `chrono::Month` enum?

it doesn't currently have this functionality and can't be accessed directly from a `NaiveDatetime`.

One option could be to implement a trait for `NaiveDatetime` that mirrors some of the time-rs api (see - https://github.com/chronotope/chrono/issues/69#issuecomment-1510506428). Though it might need to a newtype so that it can be easily used across multiple modules/crates.

pywr-core/src/recorders/aggregator.rs
```diff
@@ -7,31 +7,37 @@ pub enum AggregationFrequency {
...
             Self::Monthly => {
+                let current_month = current_date.month();
                 // Increment the year if we're in December
-                let year = if current_date.month() == Month::December {
+                let year = if current_month == 12 {
                     current_date.year() + 1
                 } else {
                     current_date.year()
                 };
+                let next_month = (current_month % 12) + 1;
```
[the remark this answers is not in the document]
Batch21: But in this case, next month needs to be in the range 1-12 not zero-based as it is used to create a date?

pywr-schema/src/model.rs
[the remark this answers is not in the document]
jetuk: We could leave it as a date that automatically gets added 00:00:00. Or make a untagged enum for a Date string or this date-time struct?

jetuk:
I wonder if it would be beneficial to have two different `TimeDomain` types? One for fixed time-step another for variable? I'm not sure how this would interact with things like the virtual storage history though.

pywr-core/src/recorders/aggregator.rs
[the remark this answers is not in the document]
jetuk: Yes, I get that. It's just the `current_month % 12` was a bit confusing as it is not the current zero-based month.

pywr-core/src/parameters/profiles/monthly.rs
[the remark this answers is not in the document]
jetuk: The function works as it is. I think leave it as is.

pywr-schema/src/model.rs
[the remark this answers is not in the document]
Batch21: I've gone with an untagged enum which seems to work.

Batch21:
@jetuk I think this is ready for you to take another look

pywr-core/src/timestep.rs
```diff
@@ -1,23 +1,61 @@
...
+    }
+}
+
+impl PywrDuration {
+    pub fn days(days: i64) -> Self {
+        Self(TimeDelta::days(days))
+    }
+
+    pub fn whole_days(&self) -> Option<i64> {
+        if self.fractional_days() % 1.0 == 0.0 {
```
jetuk, line 31: Does this pass clippy checks of comparing against floating point values?

pywr-core/src/recorders/aggregator.rs
```diff
@@ -47,7 +48,7 @@ impl AggregationFrequency {
         let mut sub_values = Vec::new();
 
         let mut current_date = value.start;
-        let end_date = value.start + value.duration;
+        let end_date = value.start + *value.duration.time_delta();
```
jetuk, line 51: One alternative  here would be to implement `Add` for `PywrDuration` with `rhs = NaiveDateTime`

pywr-core/src/models/mod.rs
```diff
@@ -16,11 +17,11 @@ impl ModelDomain {
         Self { time, scenarios }
     }
 
-    pub fn from(timestepper: Timestepper, scenario_collection: ScenarioGroupCollection) -> Self {
-        Self {
-            time: timestepper.into(),
+    pub fn from(timestepper: Timestepper, scenario_collection: ScenarioGroupCollection) -> Result<Self, PywrError> {
```
jetuk, line 20: Rename this to `try_from`

pywr-core/src/parameters/profiles/monthly.rs
```diff
@@ -5,8 +5,8 @@ use crate::state::{ParameterState, State};
 use crate::timestep::Timestep;
 use crate::PywrError;
 use std::any::Any;
-use time::util::days_in_year_month;
-use time::Date;
+
```
jetuk, line 8: Remove empty line

pywr-core/src/parameters/py.rs
```diff
@@ -120,7 +121,7 @@ impl Parameter for PyParameter {
                 py,
                 timestep.date.year(),
                 timestep.date.month() as u8,
-                timestep.date.day(),
+                timestep.date.day() as u8,
```
jetuk, line 124: I think we can use the `chrono` feature in pyo3 to pass the date directly as an argument.

pywr-core/src/parameters/py.rs
```diff
@@ -154,7 +155,7 @@ impl Parameter for PyParameter {
                     py,
                     timestep.date.year(),
                     timestep.date.month() as u8,
-                    timestep.date.day(),
+                    timestep.date.day() as u8,
```
jetuk, line 158: As above.

pywr-core/src/recorders/hdf.rs
```diff
@@ -38,8 +39,8 @@ impl Date {
         Self {
             index: ts.index,
             year: ts.date.year(),
-            month: ts.date.month().into(),
-            day: ts.date.day(),
+            month: ts.date.month() as u8,
+            day: ts.date.day() as u8,
```
jetuk, line 43: Do we need to support saving the HMS data in the HDF file?

pywr-core/src/test_utils.rs
```diff
@@ -14,20 +14,28 @@ use crate::solvers::ClpSolver;
...
+        .unwrap()
+        .and_hms_opt(0, 0, 0)
+        .unwrap();
+    let duration = TimestepDuration::Days(1);
+    Timestepper::new(start, end, duration)
 }
 
 pub fn default_time_domain() -> TimeDomain {
-    default_timestepper().into()
+    TimeDomain::try_from(default_timestepper()).unwrap()
```
jetuk, line 38: `default_timestepper().try_into().unwrap()` should also work here and not require the `TimeDomain` type to written twice.

pywr-core/src/timestep.rs
```diff
@@ -1,79 +1,169 @@
+use chrono::{Months, NaiveDateTime, TimeDelta};
+use polars::datatypes::TimeUnit;
+use polars::time::ClosedWindow;
 use pyo3::prelude::*;
 use std::ops::Add;
-use time::{Date, Duration};
+
```
jetuk, line 6: Remove empty line.

pywr-core/src/timestep.rs
```diff
@@ -1,79 +1,169 @@
...
         while current.date <= self.end {
-            let next = current + self.timestep;
+            let next = current + duration;
             timesteps.push(current);
             current = next;
         }
         timesteps
     }
+
+    fn generate_timesteps_from_frequency(&self, frequency: String) -> Result<Vec<Timestep>, PywrError> {
```
jetuk, line 122: Can we add a docstring here. Also it would be more idiomatic to use `frequency: &str`.

pywr-core/src/virtual_storage.rs
```diff
@@ -352,7 +375,8 @@ mod tests {
         let recorder = AssertionFnRecorder::new("link-1-flow", Metric::NodeOutFlow(idx), expected, None, None);
         network.add_recorder(Box::new(recorder)).unwrap();
 
-        let model = Model::new(default_timestepper().into(), network);
+        let domain = ModelDomain::try_from(default_timestepper()).unwrap();
```
jetuk, line 378: `let domain = default_timestepper().try_into().unwrap();` is slightly shorter.

pywr-python/src/lib.rs
```diff
@@ -64,9 +65,13 @@ impl Schema {
     #[new]
     fn new(title: &str, start: &PyDate, end: &PyDate) -> Self {
         // SAFETY: We know that the date & month are valid because it is a Python date.
-        let start =
-            Date::from_calendar_date(start.get_year(), start.get_month().try_into().unwrap(), start.get_day()).unwrap();
-        let end = Date::from_calendar_date(end.get_year(), end.get_month().try_into().unwrap(), end.get_day()).unwrap();
+        let start = DateType::Date(
```
jetuk, line 68: Should we support datetime in this API as well?

jetuk (changes requested):

pywr-core/src/timestep.rs
Batch21: I don't get a warning locally. Is there going to be an issue with floating point precision here? Might be better to use the `float_cmp` crate?

jetuk: I was possible expecting clippy to show this: https://rust-lang.github.io/rust-clippy/master/index.html#/float_cmp_const

However, it is part of the "restriction" group which means it should be considered case-by-case. The question is whether there's any reasonable output of `fractional_days()` which we expect to be treated as a whole day, but would not be by this test? Leap seconds? Are those still a thing?

If there's not already just add a test for this method.

jetuk: @Batch21 I did a bit of reading here. I think modulo on floats is a bit weird sometimes. It might be better to do the modulo on the `num_seconds()` as a `i64`. 

Something like this might be better?

```rust
const SECS_IN_DAY: i64 = 60 * 60 * 24;

  pub fn whole_days(&self) -> Option<i64> {
      if self.0.num_seconds() % SECS_IN_DAY == 0 {
           Some(self.0.num_days())
      } else {
           None
      }
  }

  pub fn fractional_days(&self) -> f64 {
      self.0.num_seconds() as f64 / SECS_IN_DAY as f64
  }
```

Batch21:
@jetuk I think the latest commits address all your comments

jetuk (approved):
LGTM. Good update!

PR #119, opened 2024-02-21
chore: Upgrade CLP to latest release v1.17.9

jetuk:
Some additional updates to clp-sys metadata and referencing versions.


jetuk:
Hmm, looks like setting up bindgen on Windows is a PITA. I will undo the bindgen generation and just update the static `bindings.rs` instead.

Batch21 (approved):
LGTM

PR #121, opened 2024-02-24
Added WeeklyProfile parameter

s-simoncelli:
This is the new branch implementing the `WeeklyProfile` parameter. For the previous discussion see [PR#72](https://github.com/pywr/pywr-next/pull/72)


s-simoncelli:
This also includes new changes due to the removal of the `time` crate. The parameter only relies now on `chrono` and uses the updated `Timestep `struct

jetuk (commented):
The only thing I noticed from the `chrono` changes is that the interpolation only uses the date, and will not do sub-daily interpolation. Do you think it should?

Otherwise this looks good to me. Thanks for doing the new PR.

jetuk:
Associated issue #123

PR #130, opened 2024-02-27
feat: Add StateBuilder for construction of State

jetuk:
`State::new` contained a lot of similarly typed arguments. This commit removes `State::new` and replaces it with a `StateBuilder` type that contains methods for adding optional arguments (which default to zero) and `.build()` for creating a `State`.


Batch21 (approved):
Looks good. imo the updated code is more readable.

PR #131, opened 2024-03-04
Feature/include time in profile interpolation

### LINKED ISSUE #123: Sub-daily profile interpolation
Monthly and weekly profile interpolation currently only uses the date/day part of the time-step when interpolating. For sub-daily timesteps, added in #115, this will produce the same value throughout a given day.

s-simoncelli:
This should fix #123


jetuk (approved):
LGTM

PR #136, opened 2024-03-11
feat: add pywr-schema-macro crate

Batch21:
This contains a derive macro that implements `parameters` and `parameters_mut` methods for nodes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ members = [
     "pywr-schema",
     "pywr-cli",
     "pywr-python",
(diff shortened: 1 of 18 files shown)


Batch21:
@jetuk I was able to copy over your code from the v1 schema with little modification. I've left in a TODO for parsing more complex node attribute types, such as factors and links.

jetuk:
This looks good enough to unblock #99 . 

It will need some more refinement to handle the other fields types, but that can be part of a wider improvement.

jetuk (approved):

PR #137, opened 2024-03-11
feat: Implement Parameter<usize> for PyParameter.

jetuk:
This supports using PyParameter as an index parameter with updated schema to define the return type of the Python method.


Batch21 (approved):
Worked for me when loading the custom index parameter in the more complex model we were looking at the other day.

One thing I did notice is that when building the docs the `PythonReturnType` enum is not included so its not obvious how to set return type. Might be I'm not building them correctly.

PR #138, opened 2024-03-11
refactor: Make ParameterIndex generic

jetuk:
This removes `IndexParameterIndex` (awful name!) and `MultiValueParameterIndex` in favour of a generic `ParameterIndex<T>.` Where `T` is the return type of the associated `Parameter<T>`. This ensures there is stronger type constraints to prevent using the wrong index in the wrong vec of parameters.


Batch21 (approved):
Bit annoying that those traits have to manually derived but this is definitely much cleaner overall

PR #141, opened 2024-03-12
fix: Fix SIMD imports to work with latest nightly.

jetuk:
Add nightly builds to CI.


jetuk:
The errors here are fixed in pyo3 v0.23. We won't be able to upgrade that until a Polars release is done.

jetuk:
@Batch21 this is now ready for merge if you have any time for review. Not a high priority though.

Batch21 (approved):
Given it another read through and it all looks ok. There are some changes in there that seem to be unrelated to the SIMD stuff but I think that must be due to merging main into the branch a few times?

PR #142, opened 2024-03-12
fix: Fix and upgrade highs_sys to v1.6.2

jetuk:
Restores some removed functions required for the highs solver. Adds highs feature to CI.


Batch21 (approved):
Works for me now!

PR #143, opened 2024-03-12
feat: Capture conversion errors when using convert CLI.

jetuk:
Alter the framework for the convert command to capture all the conversion errors and log them to the user.


Batch21 (approved):
Tested it on a couple of models and it works nicely.

PR #148, opened 2024-03-22
refactor: Add LoadArgs to capture schema loading arguments.

jetuk:
This reduces the number of arguments to schema loading functions, and will make maintaining those arguments easier in future.


Batch21 (approved):
lots of red!

PR #149, opened 2024-03-25
fix: Fix an issue accumulating aggregated metric values.

jetuk:
The issue was caused by the lazy evaluation of `.map()` when collecting into an `Option`. This is now done explicitly.


Batch21:
The fix looks good but the PR appears to include many of the changes from #150. Was the idea to remove these?

jetuk:
Yeah, not sure what I did wrong there. Fixed it now.

Batch21 (approved):

PR #158, opened 2024-04-10
Refactor of schema metrics

jetuk:
This unifies the schema `OutputMetric` and `DynamicFloatValue` to a single schema `Metric` enum. This simplifies some of the loading and removes the heavily nested enum. 

It also implements an `OutputMetric` in core that is used to hold additional information regarding the origin of a metric in the schema. This is useful for ensuring output data aligns with the schema model definition rather than the internal details of core.

Outstanding items:

- [ ] Do something similar with `DynamicIndexValue`
- [ ] I'm not sure I like the naming of "pywr-type" and "pywr-subtype" attributes, but naming things is hard.


pywr-schema/src/parameters/mod.rs
```diff
@@ -201,69 +204,51 @@ impl Parameter {
...
-            Self::UniformDrawdownProfile(p) => ParameterType::Parameter(p.add_to_model(network, args)?),
-            Self::Max(p) => ParameterType::Parameter(p.add_to_model(network, args)?),
-            Self::Min(p) => ParameterType::Parameter(p.add_to_model(network, args)?),
-            Self::Negative(p) => ParameterType::Parameter(p.add_to_model(network, args)?),
-            Self::Polynomial1D(p) => ParameterType::Parameter(p.add_to_model(network)?),
-            Self::ParameterThreshold(p) => ParameterType::Index(p.add_to_model(network, args)?),
-            Self::TablesArray(p) => ParameterType::Parameter(p.add_to_model(network, args)?),
+            Self::Constant(p) => pywr_core::parameters::ParameterType::Parameter(p.add_to_model(network, args)?),
+            Self::ControlCurveInterpolated(p) => {
+                pywr_core::parameters::ParameterType::Parameter(p.add_to_model(network, args)?)
```
Batch21, line 220: Is switch to the full import paths in this function just for clarity?

Batch21 (approved):
This looks good to me. I can't think of any good alternatives for the hdf attribute names that are not too verbose.

PR #163, opened 2024-04-17
feat: Add core feature to pywr-schema.

jetuk:
The "core" feature enables loading and building models with `pywr-core`. It is enabled by default. The main purpose is to reduce dependencies and build times if a user only wants to interact with the schema (i.e. not build or run a model).

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@
 
 
 <!-- ABOUT THE PROJECT -->
+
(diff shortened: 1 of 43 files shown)


Batch21 (approved):
Looks good. Only tested by running the schema test with and without the core feature but its looks like it all works.

PR #167, opened 2024-04-20
feat: Add VisitMetrics and VisitPaths traits and implementation.

jetuk:
This adds methods to the network and various components for "visiting" Metrics and Paths. These methods are useful for finding, for example, all paths in a model.

The traits are trivially implemented for common types. New derive macros replace the previous PywrNode macro. These macros generate implementations of the two traits for the struct and enum types currently used in the schema.

Outstanding items:
- [ ] Should there be separate `visit_` and `walk_` methods. Where walk would be recursive, but visit not?


Batch21 (changes requested):
This is pretty cool. From reading through and expanding some of the macros it all looks good apart form couple of minor comments,

Batch21 (approved):

PR #171, opened 2024-04-25
feat: Add --network-only argument to CLI convert command

jetuk:
This uses the update pywr-v1-schema. The error handling in the convert command is also improved.


Batch21 (approved):
Tested it on a couple of files and it worked for me. Not sure if you planned to do the new scheme release first and then update the cargo file here before merging this?

jetuk:
> Not sure if you planned to do the new scheme release first and then update the cargo file here before merging this?

Yes, I will.

PR #172, opened 2024-04-25
feat: Add functionality to export a JSON schema of PywrModel.

jetuk:
Use the schemars crate to export a JSON schema via the CLI. This needs implementing for the remaining model types.


Batch21 (approved):
I did notice when looking at this that the schemars crate it not getting a lot of maintenance. Probably something to keep an eye on.

PR #175, opened 2024-04-26
build: Build Python 3.9 ABI3 wheels.

jetuk:
Enable the "abi3-py39" feature in PyO3. This will now build ABI3 compatible wheels for Python 3.9 and above. Doing this reduces the need for separate builds for each Python version, and therefore reduces the number of CI jobs.


Batch21 (approved):
Looks good

PR #191, opened 2024-05-14
fix: schema refs to v1 params that are converted to timeseries inputs are now updated to timeseries refs

Batch21:


Batch21:
Another possible way to do this would be to extract all the Timeseries references first and pass that to the node/param conversion methods but I think that might be a messier approach as we wouldn't be able to use the `TryFrom` implementations for the conversions.

It might be worth updating `visit_metrics_mut` so that errors in this update process can be captured. What do you think?

jetuk (changes requested):
This looks good. Is it worthwhile adding a test for this kind of thing?

Batch21:
> This looks good. Is it worthwhile adding a test for this kind of thing?

done

pywr-schema/src/model.rs
```diff
@@ -1223,4 +1252,44 @@ mod core_tests {
 
         model.run::<ClpSolver>(&Default::default()).unwrap();
     }
+
+    #[test]
+    fn test_v1_conversion() {
+        let v1_str = include_str!("./test_models/timeseries_v1.json");
```
jetuk, line 1258: Could I ask we put v1 files in a sub-folder for clarity?

pywr-schema/src/model.rs
```diff
@@ -1223,4 +1252,44 @@ mod core_tests {
...
+        let v1: pywr_v1_schema::PywrModel = serde_json::from_str(v1_str).unwrap();
+
+        let (v2, errors) = PywrModel::from_v1(v1);
+
+        assert_eq!(errors.len(), 0);
+        assert_eq!(v2.network.nodes.len(), 6);
+        assert_eq!(v2.network.edges.len(), 5);
+
+        // Check that a timeseries ref has been created for a node attribute
```
jetuk, line 1267: Would it be better/easier to have an expected v2 JSON file? Load up that schema and then compare the two for equality?

pywr-schema/src/model.rs
```diff
@@ -1223,4 +1252,44 @@ mod core_tests {
...
+                _ => panic!("Expected a 'factored_flow' parameter to reference a timeseries metric"),
+            }
+        }
+
+        let current_dir = std::env::current_dir().unwrap();
+        println!("Current directory: {:?}", current_dir);
+
+        let data_path = PathBuf::from_str("src/test_models").unwrap();
+
+        let model = v2.build_model(Some(data_path.as_path()), None).unwrap();
```
jetuk, line 1292: This will require the "core" feature. I wonder if the conversion tests should run without that feature?

pywr-schema/src/model.rs
```diff
@@ -977,6 +1005,26 @@ mod tests {
...
+
+        assert_eq!(errors.len(), 0);
+
+        std::fs::write("tmp.json", serde_json::to_string_pretty(&v2).unwrap()).unwrap();
+
+        let v2_converted: serde_json::Value =
+            serde_json::from_str(&serde_json::to_string_pretty(&v2).unwrap()).unwrap();
+
+        let v2_expected: serde_json::Value =
+            serde_json::from_str(include_str!("./test_models/timeseries-converted.json")).unwrap();
```
jetuk, line 1024: I think this should sit next to the v1 JSON file: `test_models/v1/timeseries-converted.json`

jetuk:
> It might be worth updating `visit_metrics_mut` so that errors in this update process can be captured. What do you think?

I'm also not sure what you mean by this? Can you give an example?

Batch21:
> > It might be worth updating `visit_metrics_mut` so that errors in this update process can be captured. What do you think?
> 
> I'm also not sure what you mean by this? Can you give an example?

I was thinking about updating the trait method to return a `Result` so that any errors in converting a parameter ref to a timeseries ref can be added to the conversion errors Vec and returned to the user. Errors can occur if the `TimeseriesColumn` struct cannot be generated from the v1 data. I've seen a few of these because, unlike v1,  v2 requires a column name attribute for CSVs with only one column of data . Currently, in this situation, the code just bails on the conversion to a timeseries ref but I think it would be more useful to show this as an error to the user.

pywr-schema/src/model.rs
Batch21: done

jetuk (approved):

PR #194, opened 2024-06-05
Refactor parameters into General, Simple and Constant 

jetuk:
This is designed as an alternative to https://github.com/pywr/pywr-next/pull/192.

The high level idea is the solve the circular parameter issue with initial volumes by recognising that some parameters
do not vary in time and/or depend no node state.

We introduce the General, Simple and Constant parameters as new traits:

- `GeneralParameter` - rename of the current implementation.
- `SimpleParameter` - Parameters that only depend on other Simple or Constant parameters, but do vary in time.
- `ConstantParameter` - Parameters that only depend on other Constant parameters, and do *not* vary in time.

Simple parameters are evaluated before general and `.before` methods each time-step. This allows them to vary in time, but means that can't depend on network or general parameter values. This is most relevant for profiles which have no dependencies but do vary in time.

Constant parameters would be evaluated at the beginning of the simulation only. NB, this is not yet implemented.

Some parameters can potentially implement one or more of these traits. For example, the AggregatedParameter is implemented for General and Simple depending on whether it is aggregating over MetricF64 or SimpleMetricF64. This allows the user to use an aggregated parameter as a simple parameter provided it is only aggregating over simple parameters. From the schema point of view there is no difference. When added to the core model it attempts to add it as a simplified version if possible.

@Batch21 @s-simoncelli @laurenpetch any thoughts on this vs #192 ? Sorry, it's a big refactor / change this one.


Batch21 (commented):
From an initial read through, this feels like a better solution than #192 

One issue I did find when testing using the delay parameter as a simple parameter is that the `after` method is not currently called simple parameters.

pywr-core/src/parameters/mod.rs
```diff
@@ -292,6 +723,574 @@ pub trait VariableParameter<T> {
...
+                Ok(index.into())
+            }
+        }
+    }
+
+    pub fn add_simple_usize(
+        &mut self,
+        parameter: Box<dyn SimpleParameter<usize>>,
+    ) -> Result<SimpleParameterIndex<usize>, PywrError> {
+        // TODO Fix this check
```
Batch21, line 1003: Should this be fixed in this PR?

jetuk:
This has been merged with master and the above issues addressed.

Batch21 (approved):

PR #197, opened 2024-06-06
docs: Add developer guide on adding a new parameter.

jetuk:


pywr-book/src/developers-guide/adding-a-parameter.md
```diff
@@ -0,0 +1,97 @@
...
+
+To allow the parameter to be used in the model it is helpful to add a `new` function that creates a new instance of the
+parameter. This will be used by the schema to create the parameter when it is loaded from a model file.
+
+```rust,ignore
+{{#rustdoc_include ../../listings/adding-a-parameter/src/main.rs:impl-new}}
+```
+
+Finally, the minimum implementation of the `Parameter` and one of the three types of parameter compute traits should be
+added for `MaxParameter`. This trait requires the `meta` function to return the metadata for the parameter, and
```
Batch21, line 52: Should be 'These traits...' ?

Batch21 (changes requested):
One minor suggestion but otherwise I think these sections read well.

PR #205, opened 2024-06-13
feat: updates to use of durations when resampling and aligning timeseries 

Batch21:
Main change is to switch to using milliseconds as the main unit for comparison and resampling. polars is also updated to v0.40.0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ thiserror = "1.0.25"
 num = "0.4.0"
 float-cmp = "0.9.0"
 ndarray = "0.15.3"
-polars = { version = "0.39", features = ["lazy", "rows", "ndarray"] }
+polars = { version = "0.40", features = ["lazy", "rows", "ndarray"] }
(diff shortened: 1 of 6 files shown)


jetuk (commented):
This looks good. However, there are a couple of places where nanoseconds are still used.

1. There's `TimeseriesError::NoDurationNanoSeconds`  which does not appear to be used.
2. `generate_timesteps_from_frequency` uses nanoseconds. Should this use milliseconds to be consistent?

Batch21:
I've removed the unused error. 

I don't think the reference to nanoseconds in `generate_timesteps_from_frequency` can be removed completely as the polars `Duration` uses them internally to store a proportion of the stochastics. We could convert to milliseconds when creating the new datetime, but this seem a bit unnecessary:

```rust
+ TimeDelta::milliseconds(duration.nanoseconds() / NANOSECS_IN_MILLISECOND)
```

jetuk (approved):

PR #206, opened 2024-06-20
fix: Fix dependency features when building pywr-core alone.

jetuk:


Batch21 (approved):
only query is whether this change means that serde does not need to be specified in features for the chrono dependency in `pywr-schema/Cargo.toml`

PR #207, opened 2024-06-20
chore: Switch to coinclp-sys for CLP bindings

jetuk:
This replaces the local clp-sys crate and the vendored CLP source code in this repository.

Fixes #201.

PR #208, opened 2024-06-24
feat: Add additional COIN-OR bindings for Cbc.

jetuk:
This requires adding Cgl and Osi libraries. Currently these are vendored in the same way as Clp. This is an alternative to #207 which seems to have build / linking issues. The longer term plan would be to move this -sys package to separate repository and remove the submodules from here. We'll keep it here for now while we develop the CBC solver and integer variables.


Batch21 (approved):
works for me

jetuk:
Shall I add an initial CBC solver implementation here, or do that in another PR?

Batch21:
> Shall I add an initial CBC solver implementation here, or do that in another PR?

Happy either way. Would be cleaner in a new PR unless you think the implementation might required modifying the changes in this PR?

jetuk:
I do not think it will.

One issue I have immediately found is that the CBC C interface has no method for modifying matrix coefficients (unlike CLP).

PR #213, opened 2024-06-26
Add soft constraints in link node

s-simoncelli:


s-simoncelli:
If there are no problems with the implementation, I may start writing a paragraph about this in the book

jetuk:
I've only had quick read through this. My initial reaction is whether the main `min_flow` and `max_flow` values should always be hard constraints. And that a sensible configuration is for the soft flow bounds to be inside the range of `min_flow` and `max_flow`. I.e. the soft max should be lower than the hard max.

I think this would require an aggregated node to enforce the hard min & max flows across the multiple internal nodes if either soft bound is defined, and the soft max applied to the main L node.

@Batch21 thoughts?

Batch21:
Keeping the soft constraints within the bounds of `min_flow`  and `max_flow`  makes sense to me.

Could apportioning volumes to the different links be an alternative to using an agg node. Something  like this:

L_min `max_flow` = soft_min
L_max `max_flow` = max_flow - soft_max
L `max_flow` =  max_flow - soft_min - (max_flow - soft_max)

s-simoncelli:
Apologies for not reply to this sooner but it's been a busy month. Wouldn't it make more sense to have the node `L` downstream to enforce the hard min and max flows? 

image


We can then add 2 aggregated nodes:
  - Constraining  `L` and `L_min` with the `min_flow` of node `L`
  - Constraining  `L` and `L_max` with the `max_flow` of node `L`

Altought one can still set the `min_flow` of `L_min` larger than the `max_flow` of `L_max`

jetuk:
> We can then add 2 aggregated nodes:
> 
>     * Constraining  `L` and `L_min` with the `min_flow` of node `L`
>     * Constraining  `L` and `L_max` with the `max_flow` of node `L`

I'm not sure I follow this. This constraint would be redundant because it could just be applied to `L` and achieve the same thing?

I was thinking of it like this:

image

If soft min and max defined:
 - a: min_flow = 0.0, max_flow = unconstrained, cost = user's soft_max_flow_cost (i.e. cost of going above soft max)
 - b: min_flow = 0.0, max_flow = unconstrained, cost = user's standard cost
 - c: min_flow = 0.0, max_flow = user's soft_min_flow, cost = user's soft_min_flow_cost

 - agg_bc: nodes: [b, c], min_flow = 0.0, max_flow = user's soft_max_flow
 - agg_abc: nodes: [a, b, c], min_flow = user's hard min_flow, max_flow = user's hard_max_flow


if only soft min defined:
 - a: not created
 - b: min_flow = 0.0, max_flow = unconstrained, cost = user's standard cost
 - c: min_flow = 0.0, max_flow = user's soft_min_flow, cost = user's soft_min_flow_cost

 - agg_bc: nodes: [b, c], min_flow = user's hard min_flow, max_flow = user's hard_max_flow

If only soft max defined:
 - a: min_flow = 0.0, max_flow = unconstrained, cost = user's soft_max_flow_cost (i.e. cost of going above soft max)
 - b: min_flow = 0.0, max_flow = user's soft_max_flow, cost = user's standard cost
 - c: not created

 - agg_abc: nodes: [a, b], min_flow = user's hard min_flow, max_flow = user's hard_max_flow

If neither soft max or soft min defined:
 - a: not created
 - b:  min_flow = user's hard min_flow, max_flow = user's hard_max_flow, cost = user's standard cost
 - c: not created

s-simoncelli:
Hello,

I implemented the new constraints as suggested.  I think it would be nice to export the internal nodes' flow data using metrics. What do you think?

I have no idea why the pipelines fail. On my machine the project compiles

pywr-schema/src/nodes/core.rs
```diff
@@ -170,7 +332,49 @@ impl LinkNode {
...
+                    None,
+                )?;
+                network.add_aggregated_node(
+                    node_name,
+                    Self::aggregated_node_l_l_min_sub_name(),
+                    &[vec![link], vec![soft_min_node]],
+                    None,
+                )?;
+            }
+            _ => {}
```
jetuk, line 376: Minor thing for if this ever updated. Using `(None, None)` instead of the `_` would prevent this accidentally catching unused combinations in the future.

pywr-schema/src/nodes/core.rs
```diff
@@ -136,23 +136,176 @@ impl TryFrom<InputNodeV1> for InputNode {
...
+/// node `U` and downstream node `D`, will look like this on the model schematic:
+///
+/// ```svgbob
+///
+///          U         L         D
+///    - - ->*-------->*--------->*- - -
+/// ```
+///
+/// # Soft constraints
+/// This node allows setting optional maximum and minimum soft constraints which means the node's `min_flow`
```
jetuk, line 157: I think this paragraph is no longer true with how you've refactored it. The node should always stay within the min and max flow bounds now?

pywr-schema/src/nodes/core.rs
```diff
@@ -1180,6 +1514,38 @@ mod tests {
...
     }
 
+    #[cfg(feature = "core")]
+    fn link_with_soft_min_str() -> &'static str {
+        include_str!("../test_models/link_with_soft_min.json")
+    }
+
+    #[test]
+    #[cfg(feature = "core")]
+    fn test_link_with_soft_min() {
```
jetuk, line 1524: Tests look good to me. I guess we're just missing some expected output. Happy to do that in a follow-up though.

jetuk (changes requested):
This looks good. I've just noticed a possibly outdated sentence in the docstring. One other very minor comment. At some point we should provide expected output data for the test models. 

Happy to merge when the doc string is fixed.

jetuk:
> I think it would be nice to export the internal nodes' flow data using metrics. What do you think?

Agreed. What are you thinking? Some attributes that might be useful (first two more than the last two):

- `(In|Out)flowAboveSoftMax` or `SoftMaxFlowExceedance` - self explanatory
- `(In|Out)flowBelowSoftMin` or `SoftMinFlowDeficit` - A deficit to the soft min flow. 
- `(In|Out)flowAboveSoftMin` - flow less the soft min flow
- `(In|Out)flowBelowSoftMax` - flow less the soft max flow

WTW node could also implement the soft min related ones.

pywr-schema/src/nodes/core.rs
s-simoncelli: Yes that's still missing. I'll open a separate issue

s-simoncelli:
Hi, thanks for the feedback. 

I created two separate issues to address the [additional metrics](https://github.com/pywr/pywr-next/issues/262) and the [test implementation](https://github.com/pywr/pywr-next/issues/261)

jetuk (approved):

jetuk:
Thanks for implementing this!

Issue #18, opened 2023-06-03
Should each solver be a cargo feature?


Batch21:
I've been testing Pywr-next on WSL which does not support OpenCl. This has meant I've had to comment out the OpenCL solver code to avoid compilation errors. A good solution is probably to make each solver a feature.

jetuk:
Agreed, and then we should agree which are default features.

jetuk:
Only CLP is left un-gated now.

jetuk:
#468 completed this.

PR #215, opened 2024-07-02
Initial implementation of CBC solver.

jetuk:
The current CBC C interface does not support modifying constraint coefficients. Therefore this solver can not support aggregated nodes with factors at present. In theory it could support constant factors, but that is also not implemented here. 

If this initial version is OK there are two follow-ups to come:
1. Refactor #198 to work with CBC as well.
2. Use an updated version of CBC that supports constraint coefficient modification.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,16 +3,16 @@ name = "pywr"
 version = "2.0.0beta"
 description = ""
 authors = [
-    {name="James Tomlinson", email="tomo.bbe@gmail.com>"}
+    { name = "James Tomlinson", email = "tomo.bbe@gmail.com>" }
 ]
 readme = "README.md"
 requires-python = ">=3.9"
 license = "MIT OR Apache-2.0"
 dependencies = [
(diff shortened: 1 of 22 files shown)


Batch21 (changes requested):
Couple of trivial comments but otherwise this looks good to me.

Would it be worth adding the solver as a cli option in this PR? Relatedly, I do not think it is currently possible to run the feature-gated solvers via the cli. Looks like it requires the features to be duplicated in the cli toml file, unless there is a way to have shared features in a workspace?

jetuk:
> Would it be worth adding the solver as a cli option in this PR? Relatedly, I do not think it is currently possible to run the feature-gated solvers via the cli. Looks like it requires the features to be duplicated in the cli toml file, unless there is a way to have shared features in a workspace?

I'll probably do this when finishing #18. It's an issue in all of the other crates that depend on `pywr-core`.

I've addressed the comments.

Batch21 (approved):

PR #217, opened 2024-07-15
feat: Utilise constant parameters in aggregated node LP creation.

jetuk:
Where possible utilise constant parameters to try to get aggregated node factor pairs during LP creation. This makes it possible to support constant factors in LP solvers without an API for updating the constraint matrix. It should also be generally more efficient by only updating non-constant factors.


pywr-core/src/aggregated_node.rs
```diff
@@ -263,7 +337,71 @@ fn get_norm_proportional_factor_pairs(
...
+            .iter()
+            .skip(1)
+            .zip(values)
+            .map(move |(&n1, f1)| {
+                NodeConstFactorPair::new(NodeConstFactor::new(n0, None), NodeConstFactor::new(n1, f1))
+            })
+            .collect::<Vec<_>>()
+    } else {
+        // All factors are available; therefore we can calculate "f0"
+        // TODO do we need to assert that each individual factor is positive?
```
Batch21, line 377: to do in this PR? Should it be a general requirement that factors are non-negative checked in the schema?

jetuk: This has been done.

Batch21 (approved):

PR #225, opened 2024-07-23
feat: allow infer_schema_length to be specified for polars timeseries csv datasets

Batch21:


Batch21:
I had issues with a timeseries data that started with multiple zeros being inferred as an integer column. I think the safest option here is to use the full dataset to infer types, unless we make it something that can be specified in the model definition.

jetuk:
Will this make parsing CSV files very slow?

Batch21:
> Will this make parsing CSV files very slow?

Yeah, I think it would. I've updated the PR so users can specify the number of lines to use to infer types for each dataset.

jetuk (approved):

PR #234, opened 2024-07-30
feat: Add a parent namespace to Parameters.

### LINKED ISSUE #179: Add parameter name spaces
For cases where a parameter needs to be added as child of another parameter or node as part of "building" a model from a schema. The current approach requires making a unique global name which has issues with risking conflict and create horrible names. 

We currently model this in nodes using "sub-names"; an optional unique identifier in the namespace of the parent node name. The same idea could work with parameters. 

However, I wonder if it would make more sense to have an optional parent (and whether the nodes should be re-modelled in this manner also).

See also discussion in #107

jetuk:
Core Parameters can now be identified with an optional parent string. This is helpful for creating Parameters in compound nodes without risking naming conflicts with the user's own parameter names.

Fixes #179.

### DIFF
--- a/pywr-schema/src/timeseries/mod.rs
+++ b/pywr-schema/src/timeseries/mod.rs
@@ -12,6 +12,7 @@ use ndarray::Array2;
 use polars::error::PolarsError;
 #[cfg(feature = "core")]
 use polars::prelude::{DataFrame, DataType::Float64, Float64Type, IndexOrder};
+use pywr_core::parameters::ParameterName;
 #[cfg(feature = "core")]
 use pywr_core::{
     models::ModelDomain,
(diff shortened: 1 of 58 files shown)


pywr-schema/src/timeseries/mod.rs
```diff
@@ -207,13 +208,13 @@ impl LoadedTimeseriesCollection {
             .ok_or(TimeseriesError::TimeseriesNotFound(name.to_string()))?;
 
         let array: Array2<f64> = df.to_ndarray::<Float64Type>(IndexOrder::default()).unwrap();
-        let name = format!("timeseries.{}_{}", name, scenario);
```
jetuk, line 210: Was this "timeseries" prefix required for some reason?

Batch21: only to miminise the risk of the name clashing with an existing non-timeseries parameter

Batch21 (approved):
LGTM

PR #236, opened 2024-08-05
ci: Use one test thread on Windows CI.

jetuk:


jetuk:
This might fix the Windows tests. However, I wonder if it's because we've got some multi-threaded deadlock issue.

Batch21:
> This might fix the Windows tests. However, I wonder if it's because we've got some multi-threaded deadlock issue.

The only issue I've had running the tests locally on windows is with the doc tests and I think that is due to them using so much memory.

Batch21 (approved):

PR #239, opened 2024-08-06
feat: add filter options for metric sets

Batch21:
These provide a convienient way to record metrics for all nodes and/or parameters in a model.

### DIFF
--- a/pywr-schema/src/metric_sets/mod.rs
+++ b/pywr-schema/src/metric_sets/mod.rs
@@ -81,6 +81,46 @@ impl From<MetricAggregator> for pywr_core::recorders::Aggregator {
     }
 }
 
+#[derive(Deserialize, Serialize, Clone, JsonSchema, Default)]
+struct MetricSetFilters {
+    #[serde(default)]
+    all_nodes: bool,
+    #[serde(default)]
+    all_parameters: bool,
+}
+
+#[cfg(feature = "core")]
+impl MetricSetFilters {
+    fn create_metrics(&self, args: &LoadArgs) -> Option<Vec<Metric>> {
+        use crate::metric::{NodeReference, ParameterReference};
+
+        if !self.all_nodes && !self.all_parameters {
+            return None;
+        }
+
+        let mut metrics = vec![];
+
+        if self.all_nodes {
+            for node in args.schema.nodes.iter() {
+                metrics.push(Metric::Node(NodeReference::new(node.name().to_string(), None)));
+            }
+        }
+
+        if self.all_parameters {
+            if let Some(parameters) = args.schema.parameters.as_ref() {
+                for parameter in parameters.iter() {
+                    metrics.push(Metric::Parameter(ParameterReference::new(
+                        parameter.name().to_string(),
+                        None,
+                    )));
+                }
+            }
+        }
+
+        Some(metrics)
+    }
+}
+
 /// A set of metrics that can be output from a model run.
 ///
 /// A metric set can optionally have an aggregator, which will apply an aggregation function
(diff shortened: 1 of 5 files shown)


pywr-schema/src/metric_sets/mod.rs
```diff
@@ -81,6 +81,46 @@ impl From<MetricAggregator> for pywr_core::recorders::Aggregator {
...
+                metrics.push(Metric::Node(NodeReference::new(node.name().to_string(), None)));
+            }
+        }
+
+        if self.all_parameters {
+            if let Some(parameters) = args.schema.parameters.as_ref() {
+                for parameter in parameters.iter() {
+                    metrics.push(Metric::Parameter(ParameterReference::new(
+                        parameter.name().to_string(),
+                        None,
```
jetuk, line 114: I think this might not work if the parameter is a multi-valued one.

pywr-schema/src/metric_sets/mod.rs
```diff
@@ -91,18 +131,46 @@ pub struct MetricSet {
     pub name: String,
     pub metrics: Vec<Metric>,
     pub aggregator: Option<MetricAggregator>,
+    #[serde(default)]
+    filters: MetricSetFilters,
```
jetuk, line 135: This should be `pub` as it is part of the schema. It will require making `MetricSetFilters` public also.

We should also add to the docstring that these filters append to `metrics`, but will avoid creating duplicates. 

Should the `metrics` field be `Option<_>` now?

pywr-schema/src/metric_sets/mod.rs
```diff
@@ -91,18 +131,46 @@ pub struct MetricSet {
...
     #[cfg(feature = "core")]
     pub fn add_to_model(&self, network: &mut pywr_core::network::Network, args: &LoadArgs) -> Result<(), SchemaError> {
-        // Convert the schema representation to internal metrics.
-        let metrics: Vec<_> = self
+        use pywr_core::recorders::OutputMetric;
+
+        let mut metrics: Vec<OutputMetric> = self
             .metrics
             .iter()
+            //.chain(additional_metric.iter())
```
jetuk, line 146: Remove

pywr-schema/src/metric_sets/mod.rs
```diff
@@ -91,18 +131,46 @@ pub struct MetricSet {
...
             .map(|m| m.load_as_output(network, args))
             .collect::<Result<_, _>>()?;
 
+        if let Some(additional_metrics) = self.filters.create_metrics(args) {
+            for m in additional_metrics.iter() {
```
jetuk, line 151: If `OutputMetric` derived `PartialEq` you can simplify this to:

```rust
            for m in additional_metrics.iter() {
                let o = m.load_as_output(network, args)?;
                if !metrics.contains(&o) {
                    metrics.push(o);
                }
            }
```

jetuk: Using a HashSet would be quicker, but requires `Hash` to be derived on `OutputMetric` and will also lose the ordering.

jetuk (changes requested):

pywr-schema/src/metric_sets/mod.rs
Batch21: It doesn't. I've added a temp fix which means that a Python parameter with a dict return value is skipped by the metric set filter. I think this is currently the only multi-valued parameter?

A better solution might be to create a separate name-spaced output metric for each return value of a multi-valued parameter. This might require the Python parameter to be updated so that these values are defined upfront by the user.

pywr-schema/src/metric_sets/mod.rs
Batch21: done

pywr-schema/src/metric_sets/mod.rs
jetuk: It probably makes sense that each parameter struct in the schema has a method by which it can return a `ParameterReference` to itself.

jetuk (changes requested):
This looks good to me.

The schema-only build needs fixing by feature gating a couple of imports.

Can we also open an issue to resolve the parameter reference generation and remove the fragile filter here?

pywr-schema/src/metric_sets/mod.rs
Batch21: Should the nodes all have a similar method as well?

jetuk: Yes, probably.

jetuk (approved):

PR #240, opened 2024-08-07
feat: Initial support for Pandas timeseries backend.

jetuk:
Basic support for loading dataframes using Pandas. This is implemented using a Python callback function that loads the dataframe using Pandas, and then converts it to Polars.


pywr-schema/src/timeseries/pandas.rs
```diff
@@ -0,0 +1,107 @@
...
+                    .call((fp, index_col), kwargs.as_ref())?
+                    .extract()?;
+
+                Ok(df)
+            })
+            .expect("Failed to load Pandas dataset");
+
+            let mut df = df.0;
+
+            println!("Loaded Pandas dataset: {}", df);
```
Batch21, line 93: remove print statement?

Batch21 (changes requested):
Looks good though to get it to work with my venv I had to add the path to the site-packages folder as suggested here - https://github.com/PyO3/pyo3/discussions/3726 It looks like pyo3 doesn't yet have great support for virtual environments.

jetuk:
This worked fine for me when running the tests with my virtual environment activated. However, I appreciate that may not be working for everyone. The CI has the same issue with the Rust workflows not having the correct Python libraries available. 

If this is to go ahead we might need to merge the workflows or only run some of the tests in the Python workflows.

One general option is to add a "python" feature to the schema crate that would include things like this Pandas support.

jetuk:
What do you think is the best way forward with this?

jetuk:
I've featured gated the Rust tests that need a Python environment to work. They are not currently run in the CI.

PR #246, opened 2024-09-04
chore(deps): Switch to hdf5_metno fork.

jetuk:


jetuk:
This is a bit of a mess of ndarray versions. It'll be easier once Polars upgrades to ndarray v0.16.

Batch21 (approved):

PR #248, opened 2024-09-07
feat: Implement deny_unknown_fields for many types.

### LINKED ISSUE #165: Add #[deny_unknown_fields] to all schema structs.
This will reduce errors by having serialisation fail on unknown fields. I think this is probably the preferred behaviour.

jetuk:
Enforce `deny_unknown_fields` in parameters and node schemas. This will help finding schema errors and avoid silently ignoring incorrect or misnamed fields. This has required not flattening the meta objects in the JSON schema. Therefore, the schema has changed with the meta data now nested in a "meta" field. The meta objects do not implement `deny_unknown_fields` which allows this object to contain open ended data that might be useful for other tools.

Fixes #165.

### DIFF
--- a/pywr-python/tests/models/aggregated-node1/model.json
+++ b/pywr-python/tests/models/aggregated-node1/model.json
@@ -10,7 +10,9 @@
   "network": {
     "nodes": [
       {
-        "name": "input1",
+        "meta": {
+          "name": "input1"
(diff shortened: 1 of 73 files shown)


Batch21 (approved):
Looks good

Issue #186, opened 2024-05-12
Add LossLink type functionality to a river


jetuk:
To more easily model river losses. Some options here:

- [ ] Loss is flow arbitrary parameter
- [ ] Loss is a proportion of the river flow (this is probably the place to start)
- [ ] User enters flow-loss relationship which is linearly interpolation

s-simoncelli:
Implemented in #258

PR #258, opened 2024-09-30
Added proportional and interpolated losses in RiverNode

s-simoncelli:
This addresses #186 and implements a loss on the river node using a proportional factor or a piecewise linear interpolation of loss data


pywr-schema/src/nodes/river.rs
```diff
@@ -31,17 +75,82 @@ impl RiverNode {
...
     }
+
     pub fn add_to_model(&self, network: &mut pywr_core::network::Network) -> Result<(), SchemaError> {
         network.add_link_node(self.meta.name.as_str(), None)?;
+
+        // add output node and edge
+        if self.loss.is_some() {
+            network.add_output_node(self.meta.name.as_str(), Self::loss_node_sub_name())?;
+
+            let river = network.get_node_index_by_name(self.meta.name.as_str(), None)?;
```
jetuk, line 98: You could get these indices from the calls to `add_link_node` and `add_output_node` respectively.

pywr-schema/src/nodes/river.rs
```diff
@@ -31,17 +75,82 @@ impl RiverNode {
...
+            let loss = network.get_node_index_by_name(self.meta.name.as_str(), Self::loss_node_sub_name())?;
+            network.connect_nodes(river, loss)?;
+        }
+
+        Ok(())
+    }
+
+    pub fn set_constraints(
```
jetuk, line 106: This function needs calling in `Node::set_constraints`

pywr-schema/src/nodes/river.rs
```diff
@@ -31,17 +75,82 @@ impl RiverNode {
...
+        &self,
+        network: &mut pywr_core::network::Network,
+        args: &LoadArgs,
+    ) -> Result<(), SchemaError> {
+        let river_node = network.get_node_index_by_name(self.meta.name.as_str(), None)?;
+        let current_flow_metric = MetricF64::NodeInFlow(river_node);
+
+        if let Some(loss) = &self.loss {
+            // add the flow
+            let loss_metric = match &loss.data {
```
jetuk, line 116: One issue with this approach is that the proportion is going to be a day out. The metric will always be the previous flow. That's fine, but probably needs to be made clear. 

I wonder if we need an option to use an `AggregatedNode` to do proportional losses within the LP (I.e. the same time-step). However, the interpolation is not possible then.

pywr-schema/src/nodes/river.rs
```diff
@@ -31,17 +75,82 @@ impl RiverNode {
...
+                    let loss_idx = network.add_parameter(Box::new(interpolated_loss_parameter))?;
+                    let interpolated_loss_metric: MetricF64 = loss_idx.into();
+                    interpolated_loss_metric
+                }
+            };
+
+            network.set_node_max_flow(self.meta.name.as_str(), Self::loss_node_sub_name(), Some(loss_metric))?;
+
+            // add the optional cost
+            if let Some(cost) = &loss.cost {
```
jetuk, line 146: With an `AggregatedNode` the cost would not be required.

pywr-schema/src/nodes/river.rs
jetuk: Also, this would align with how the WTW node does things.

jetuk (commented):
I've had a quick read. Just some comments about using parameters for the losses here.

s-simoncelli:
The loss is now calculated as in the LossLink

jetuk (approved):
LGTM

Would be nice to have a quick test model of this though.

jetuk (approved):

PR #264, opened 2024-10-07
feat: Data table improvements

jetuk:
- More informative error message when failing to load a table reference.
- Implement loading array data with column headers.


Batch21 (approved):
LGTM. At some point it would be good to add some docstrings to this module and some more extensive tests. I don't think any of the current test models load data from tables?

jetuk:
Agreed. I will add an issue for that.

PR #266, opened 2024-10-08
fix: Fix handling of zero loss in LossLink and WTW nodes.

jetuk:
When a constant zero loss is provided to LossLink or WTW nodes then no relationship is created on the internal aggregated node. This left the added Output node unconstrained. This fix ensures that the Output node is separately constrained to zero max_flow when there is zero loss.


Batch21 (approved):
Looks good!

PR #267, opened 2024-10-11
fix: Clamp storage volumes to be within min & max volume ranges.

jetuk:
This accounts for small floating point rounding errors and tolerances from LP results. In some cases the resulting volume will be a rounding error above or below the bounds.

This also replicates behaviour from v1.


jetuk:
I've updated this slightly to panic if the volume is too far out of range.

PR #271, opened 2024-10-17
feat: Add Edge metric

jetuk:
A new schema and core level metric for returning edge flows. The schema level metric returns core a metric that covers the case where multiple core edges are created from a single schema edge.

Includes some additional tests for picewise-link1.json. This covers the case described above.


Batch21 (approved):
This looks good to me

PR #274, opened 2024-10-23
fix: Correct some issues with VirtualStorage node.

jetuk:
- Remove unused field `flow_constraints`.
- Make flow factors required and update the builder.
- Builder now provides unit factors if none are given.
- Improve API to return slice reference and iterator (removes some allocations).


Batch21 (approved):
LGTM

Issue #249, opened 2024-09-16
Should inline parameters be removed?


jetuk:
They are a source of headache in loading and schema interpretation due to the recursion. They can also make models difficult to read. The new schema has already required them to be named (a change from v1.x). 

I wonder if we should just remove them entirely? 

The case of a constant parameter has been superseded by the separate constant metric variant (i.e. not a parameter). Therefore constants would still be allowed inline. Trivial control curves (e.g. licence target) might be one area where they are used, but if it is a common pattern we should just create a node type to do that for the user.

Unless there are strong objections this is leaning to a "yes" at the moment.

Batch21:
My vote is "yes". I've tended to avoid inline parameters as much as possible in recent years due to them making models less readable

s-simoncelli:
I tend to use inline parameters most of the time. If they are only applied to one component (for example a node), I can immediately see how the component is setup without scrolling to another part of the JSON file. I however understand that it would make things much easier to get rid of them.

I don't really have a definitive opinion on this 😄

jetuk:
How about some design where we remove inline parameters, but provide a local space for parameters to be defined in a node. I.e. a "parameters" list on each node, that when added to the model are owned/parented by that node. This would allow local parameters to be defined on the node, but without the complexity of supporting the recursion.

jetuk:
Here's an example with two nodes referring to local parameters of the same name.

```json
[
  {
    "meta": {
      "name": "supply1"
    },
    "type": "Input",
    "max_flow": {
      "type": "LocalParameter",
      "name": "max_flow"
    },
    "parameters": [
      {
        "meta": {
          "name": "max_flow"
        },
        "type": "Constant",
        "value": 10.0
      }
    ]
  },
  {
    "meta": {
      "name": "supply2"
    },
    "type": "Input",
    "max_flow": {
      "type": "LocalParameter",
      "name": "max_flow"
    },
    "parameters": [
      {
        "meta": {
          "name": "max_flow"
        },
        "type": "Constant",
        "value": 12.0
      }
    ]
  }
]
```

jetuk:
Completed in #280 and #298.

PR #280, opened 2024-11-04
feat: Remove inline parameters.

jetuk:
This is the first step in resolving #249. Remove inline-parameter support in the v2 schema. This requires some more complicated logic for conversion from the v1 schema. The conversion method is refactored and updated to accommodate moving v1 inline parameters to the global parameter list in v2. The same method is also used for conversion to timeseries.


Batch21 (changes requested):
Given this a read through and other than a couple of minor comments it looks good to go. Was there anything else to do on this PR?

Batch21 (approved):

PR #287, opened 2024-11-13
refactor(tests): Refactor tests into integration tests.

jetuk:
Move the test models into proper `tests` folder as proper integration tests. Test functions are generated using a macro in a consistent manner. These tests no also read JSON models and CSV results from file rather than including them in the test binary.

Where original unit tests had assertion recorders to test outputs these have been converted to expected CSV output files.


pywr-schema/tests/test_models.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+}
+
+model_tests! {
+    test_simple1: ("simple1.json", vec![], vec![]),
+    test_csv1: ("csv1.json", vec!["csv1-outputs-long.csv", "csv1-outputs-wide.csv"], vec![]),
+    test_csv2: ("csv2.json", vec!["csv2-outputs-long.csv", "csv2-outputs-wide.csv"], vec![]),
+    test_csv3: ("csv3.json", vec!["csv3-outputs-long.csv"], vec![]),
+    test_hdf1: ("hdf1.json", vec![], vec![]), // TODO asserting h5 results not possible with this framework
+    test_memory1: ("memory1.json", vec![], vec![]),  // TODO asserting memory results not possible with this framework
```
Batch21, line 45: I assume these will need custom testing functions in this script and this will be the case for any tests that cannot be done by checking csv outputs?

jetuk: In this case I left the unit tests in the modules, but included them here to test deserialisation.

Batch21 (approved):

PR #298, opened 2024-11-26
feat: Node-local namespace for parameters.

jetuk:
This allows parameters to be defined within the Node schema. These parameters are given a parent namespace of the node's name. And can only be accessed by metrics of that node.

This provides an alternative to the removed inline parameters, allowing modellers to keep node-relevant parameters part of the node's definition.

Second part of #249.


jetuk:
One thing this does not do is convert inline parameters on nodes to node local parameters.

Batch21 (approved):
This looks good.

PR #305, opened 2024-12-06
fix: Fix converting dataframe parameters which reference tables.

jetuk:
Handle conversion of v1.x dataframe parameters that reference tables.


Batch21 (approved):
Looks good!

PR #317, opened 2024-12-17
feat: Replace DynamicIndexValue with IndexMetric 

### LINKED ISSUE #291: Index (usize) metric improvements.
The current `DynamicIndexValue` is the usize equivalent of `Metric`. I think the former should brought more inline with the later setup. This would remove the nested enums in favour of a more "flat" approach. Tasks:

- [ ] Use `ParameterReference` instead of `String`
- [ ] Support index (usize) timeseries
- [ ] Support indes (usize) inter-network transfers

jetuk:
`IndexMetric` mirrors `Metric`, but for integers.

Also adds support for:
- Integer arrays in core.
- `MultiValue<usize>` parameter return type.

Fixes #291

### DIFF
--- a/pywr-schema/src/metric.rs
+++ b/pywr-schema/src/metric.rs
@@ -17,7 +17,10 @@ use crate::v1::{ConversionData, TryFromV1, TryIntoV2};
 use crate::ConversionError;
 #[cfg(feature = "core")]
 use pywr_core::{
-    metric::MetricF64, models::MultiNetworkTransferIndex, parameters::ParameterName, recorders::OutputMetric,
+    metric::{MetricF64, MetricUsize},
+    models::MultiNetworkTransferIndex,
+    parameters::ParameterName,
+    recorders::OutputMetric,
 };
 use pywr_schema_macros::PywrVisitAll;
 use pywr_v1_schema::parameters::ParameterValue as ParameterValueV1;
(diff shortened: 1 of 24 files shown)


pywr-schema/src/metric.rs
```diff
@@ -401,3 +463,143 @@ impl EdgeReference {
...
+        v1: ParameterValueV1,
+        parent_node: Option<&str>,
+        conversion_data: &mut ConversionData,
+    ) -> Result<Self, Self::Error> {
+        let p = match v1 {
+            // There was no such thing as s constant index in Pywr v1
+            // TODO this could print a warning and do a cast to usize instead.
+            ParameterValueV1::Constant(value) => {
+                // Check if the value is not a whole non-negative number
+                if value.fract() != 0.0 && value >= 0.0 {
```
Batch21, line 567: Should this be `value.fract() != 0.0 || value < 0.0`

Batch21 (commented):
One query but otherwise looks good.

I've haven't tested the methods to load usize column and df timeseries data though they look like they should work. We should probably add a test model that does this at some point.

I guess its a separate PR to add `IndexMetric` support for `MetricSets`?

pywr-schema/src/metric.rs
jetuk: Fixed and added a test for it.

Batch21 (approved):

PR #318, opened 2024-12-19
fix: Try to convert RiverSplitWithGauge factors correctly.

### LINKED ISSUE #241: Conversion of factors in RiverSplitWithGaugeNode is incorrect
The implementation `TryFrom<RiverSplitWithGaugeNodeV1> for RiverSplitWithGaugeNode` is not correct. The representation of the factors is the not the same between v1.x and v2.x of Pywr. v2.x requires factors as a proportion of the main flow. Whereas v1.x requires specifying the factors as ratios.

https://github.com/pywr/pywr-next/blob/745d6d0571e70335e7c33baf10ba12958e0ff771/pywr-schema/src/nodes/river_split_with_gauge.rs#L220

Possible solutions:
- Return a `ConversionError` stating these factors cannot be converted automatically,
- Update the node to align more closely with v1.x and allow the conversion to be made, or
- Add interim parameters to calculate the proportion correctly.

jetuk:
Add a conversion function for the factors on RiverSplitWithGauge. The factors have changed in this implementation and require conversion from ratios to proportions. This is only possible if they existing factors are all constants.

Fixes #241.

### DIFF
--- a/pywr-schema/src/error.rs
+++ b/pywr-schema/src/error.rs
@@ -126,4 +126,6 @@ pub enum ConversionError {
     TableRef { attr: String, name: String, error: String },
     #[error("Unrecognised type: {ty}")]
     UnrecognisedType { ty: String },
+    #[error("Non-constant value cannot be converted automatically.")]
+    NonConstantValue {},
 }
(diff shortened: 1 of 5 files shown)


Batch21 (approved):
Looks good

PR #321, opened 2024-12-29
fix: account for leap day when indexing daily profiles

Batch21:
This matches the behaviour in v1

### DIFF
--- a/pywr-core/src/timestep.rs
+++ b/pywr-core/src/timestep.rs
@@ -125,6 +125,10 @@ impl Timestep {
         self.duration.fractional_days()
     }
 
+    /// Returns the day of the year index of the timestep.
+    ///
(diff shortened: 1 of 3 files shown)


pywr-core/src/timestep.rs
```diff
@@ -118,6 +124,14 @@ impl Timestep {
     pub(crate) fn days(&self) -> f64 {
         self.duration.fractional_days()
     }
+
+    pub fn day_of_year_index(&self) -> usize {
```
jetuk, line 128: Add a docstring explaining what this returns and how it handles leap years.

jetuk (changes requested):
There's a couple of unused imports to remove to satisfy Clippy.

pywr-core/src/timestep.rs
Batch21: Done

jetuk (approved):

PR #323, opened 2025-01-03
feat: Use u64 instead of usize for index parameters.

jetuk:
U64 is more appropriate for data that should be the same across multiple platforms. `usize` is platform specific.

This does not change the use of `NonZeroUsize` for the rolling window size in virtual storage nodes. Ultimately this values is used to size an array and that maximum size will be platform dependent.

Currently there is some casting to `usize`. A separate update could warn on the `cast_lossless` lint and remove the use of `as` throughout.

One outstanding query is whether this should actually be i64 to support negative values?

Issue #291.

### DIFF
--- a/pywr-core/src/parameters/mod.rs
+++ b/pywr-core/src/parameters/mod.rs
@@ -405,22 +405,22 @@ impl ParameterStates {
     ) -> Option<&mut Option<Box<dyn ParameterState>>> {
         self.constant.f64.get_mut(*index.deref())
     }
-    pub fn get_general_mut_usize_state(
+    pub fn get_general_mut_u64_state(
         &mut self,
-        index: GeneralParameterIndex<usize>,
+        index: GeneralParameterIndex<u64>,
     ) -> Option<&mut Option<Box<dyn ParameterState>>> {
         self.general.usize.get_mut(*index.deref())
     }
 
-    pub fn get_simple_mut_usize_state(
+    pub fn get_simple_mut_u64_state(
         &mut self,
-        index: SimpleParameterIndex<usize>,
+        index: SimpleParameterIndex<u64>,
     ) -> Option<&mut Option<Box<dyn ParameterState>>> {
         self.simple.usize.get_mut(*index.deref())
     }
-    pub fn get_const_mut_usize_state(
+    pub fn get_const_mut_u64_state(
         &mut self,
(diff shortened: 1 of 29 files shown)


pywr-core/src/parameters/mod.rs
```diff
@@ -763,19 +763,19 @@ pub struct ParameterCollectionSize {
 #[derive(Default)]
 pub struct ParameterCollection {
     constant_f64: Vec<Box<dyn ConstParameter<f64>>>,
-    constant_usize: Vec<Box<dyn ConstParameter<usize>>>,
+    constant_usize: Vec<Box<dyn ConstParameter<u64>>>,
```
Batch21, line 766: should the attributes be renamed to  `constant_u64`, `simple_u64`, and `simple_u64`?

pywr-core/src/parameters/mod.rs
```diff
@@ -1067,7 +1064,7 @@ impl ParameterCollection {
             .map(|p| p.as_parameter())
     }
 
-    pub fn get_usize_index_by_name(&self, name: &ParameterName) -> Option<ParameterIndex<usize>> {
+    pub fn get_usize_index_by_name(&self, name: &ParameterName) -> Option<ParameterIndex<u64>> {
```
Batch21, line 1067: update function name

pywr-schema/src/metric.rs
```diff
@@ -276,7 +276,7 @@ impl NodeReference {
         &self,
         _network: &mut pywr_core::network::Network,
         args: &LoadArgs,
-    ) -> Result<MetricUsize, SchemaError> {
+    ) -> Result<MetricU64, SchemaError> {
```
Batch21, line 279: function should be renamed `load_u64`?

pywr-schema/src/metric.rs
```diff
@@ -418,7 +418,7 @@ impl ParameterReference {
         &self,
         network: &mut pywr_core::network::Network,
         parent: Option<&str>,
-    ) -> Result<MetricUsize, SchemaError> {
+    ) -> Result<MetricU64, SchemaError> {
```
Batch21, line 421: function should be renamed load_u64?

Batch21 (changes requested):
A few renaming suggestions

Batch21 (approved):

PR #326, opened 2025-01-05
feat: Add pyo3 feature for Python support.

jetuk:
Adds a new default feature, `pyo3`, for enabling Python support. It is useful to feature gate PyO3 functionality to allow pywr-schema to built on a wider range of platforms (e.g. WASM).

This also updates the Linux CI to use cargo hack to check every feature combination is functional.


Batch21 (approved):
LGTM

PR #345, opened 2025-02-24
Scenario subsets

jetuk:
Below is the main contribution, but some earlier commits for tidying and refactoring are here also. Can squash or keep as required. 

This adds support for running subsets of the global scenario
domain. It implements slicing or user defined combinations. This
comes with an update to the JSON schema to put both "groups" and
"combinations" keys in the "scenarios" object. This is a revision
from v1 where "scenario_combinations" was a top-level key in the
JSON document.

In core the scenario index now stores the simulation based
indices and original schema (global) indices and labels. This
is useful for providing the correct global context in output
files.

Those parameters which use scenario indexing for looking up
data now do so using the simulation (i.e. subset) indices, and
therefore assume the data (e.g. a 2D array) is correctly sliced
for the given domain. This has been done for the parameters
that exist, but may need some additional work to be helpful
for custom parameters.

Updated CSV and HDF output formats now include concatenated
label from the global context.

New tests for slicing and combinations.


Batch21 (commented):
Just been reading through this and the implementation looks good. However I wonder now if it might be better to use a different approach to V1. Instead of having a top level combinations attribute, each scenario group could allow a list of members to be specified. This could be part of an enum with the group slice:

``` rust
enum ScenarioGroupSubset {
    Slice(usize, usize),
    Members(Vec<Usize>),
}
``` 
This would arguably more intuitive to use and more flexible as it would allow slices and member lists to be used together.

we might have already discussed this approach?

jetuk:
> This would arguably more intuitive to use and more flexible as it would allow slices and member lists to be used together.

Interesting idea. The members on groups would create the full combination of scenarios across all groups. That's slightly different to an arbtirary list of combinations. Are you saying we would keep that as well?

Batch21:
I was imagining this would be a replacement for the combinations list. I think it would generally be more concise to specify the members individually and you also wouldn't need to worry about scenario order.

jetuk:
I've finally finished this. It should be a lot more flexible now. See the book content.

pywr-schema/src/timeseries/mod.rs
```diff
@@ -339,6 +350,22 @@ impl LoadedTimeseriesCollection {
...
+#[cfg(feature = "core")]
+pub fn subset_array2<T>(array: &Array2<T>, subset: &[usize]) -> Array2<T>
+where
+    T: Copy,
+{
+    // Slice the array to only include the columns that are used in the scenario
+    let slices = subset.iter().map(|c| array.slice(s![.., *c])).collect::<Vec<_>>();
+    // Stack the slices to create a new array; this should be infallible because
+    // the slices are all the same length.
+    ndarray::stack(ndarray::Axis(1), &slices).unwrap()
```
Batch21, line 366: handle error here?

Batch21 (changes requested):
These are some nice improvements over v1 and the book chapter is very useful. Some very minor comments/suggestions from reading through it.

jetuk:
I just realised I addressed these comments but never pushed the commit.

Batch21 (approved):
Changes look good. What is causing the CI failure?

PR #352, opened 2025-03-01
fix: rechunk polars df columns before conversion to ndarrys

Batch21:
Avoids a 'chunked array is not contiguous' error during conversion.


Batch21:
something that should always be done according to this issue - https://github.com/pola-rs/polars/issues/19918

jetuk:
This looks fine, but how do we triggered the need for it / test it? I.e. are these definitely the only places we need to do this?

jetuk (approved):

jetuk (approved):

PR #358, opened 2025-03-27
Minor bug fixes and improvements

s-simoncelli:


pywr-schema/src/timeseries/mod.rs
```diff
@@ -461,6 +461,13 @@ impl TryFromV1<DataFrameParameterV1> for ConvertedTimeseriesReference {
                 Some(v) => v.as_str().map(|s| s.to_string()),
                 None => None,
             };
+            // remove the parse_dates for CSV files as this is already passed to read_csv in
```
jetuk, line 464: > ... already passed to read_csv ...

Would it be less surprising if this was not done automatically?

jetuk (approved):

pywr-schema/src/timeseries/mod.rs
jetuk: @Batch21 thoughts on this before I merge these changes.

Batch21: Maybe less suprising but you're always going to need to parse dates for timeseries and doing it automatically will make model files a little more concise.

PR #385, opened 2025-05-28
feat: Add RollingParameter.

jetuk:
Add a rolling parameter. Add a generic rolling parametr to core for use with f64 and u64 values. These two variants are also added to the schema.

Some additional trait implementations and test utility functions have been added to make this work. These mirror the f64 implementations.

One indexing bug is also corrected.


jetuk:
Thanks. I've addressed this comments.

Batch21 (approved):

PR #403, opened 2025-07-15
refactor: Refactor Timeseris into a single enum.

jetuk:
Removes `TimeseriesProvider` and makes a single top-level enum
 similar to the other components (e.g. `Node`, `Parameter`).


Batch21 (approved):
Looks good

PR #398, opened 2025-07-01
refactor: Remove giant PywrError in pywr-core.

jetuk:
Split this error into multiple modular errors. There's now error associated with more individual components, and sometimes actions of those components. This should hopefull allow for some more nuanced error handling.

Also, general favour returning `Option<T>` instead of `Result` for the `get_xxx` methods on the network, models, etc. This more closely aligns with Rust convention on types like `Vec`, etc.

PR #404, opened 2025-07-15
chore: Some Python maintenance.

jetuk:
- Remove pre-commit hooks (they are annoying).
- Tidy-up dev dependencies in pyproject.toml.
- Use uv for Python CI.
- Use ruff for Python formatting, and add to CI.
- Move CI to Python 3.10.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Check formatting
         run: uv run ruff format --check
       - name: Run tests
-        run: uv run pytest tests
+        run: uv run pytest pywr-python/tests
       - name: Build the wheels
         run: uv build
       - name: Upload wheels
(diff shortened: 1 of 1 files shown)


.github/workflows/python.yml
```diff
@@ -10,244 +7,57 @@ on:
...
-      - name: Install built wheel
-        run: |
-          pip install pytest h5py pandas polars pyarrow click
-          pip install pywr --no-index --find-links dist --force-reinstall --no-deps
-          python -c "import pywr"
-          python -m pytest
+      - name: Install the project
+        run: uv sync --locked --all-extras --dev
+      - name: Check formatting
+        run: uv run ruff format --check
```
Batch21, line 26: Also check linting - `uv run ruff check .` ?

Batch21 (changes requested):

.github/workflows/python.yml
jetuk: To be more specific, I have some larger updates to the Python extension in a branch that is waiting on #398 to be merged.

PR #410, opened 2025-07-22
Several Python extension improvements

jetuk:
The main change here is a refactor of the Python custom parameter interface. This now uses a new `ParameterInfo` object that instead of 4 separate arguments. This struct is exposed as a Python object with several methods to access data, etc. This has required exposing `Timestep` and `ScenarioIndex` as Python objects. 

In addition there are several updates to the error handling of the Python parameter. This should help debugging user's custom parameters. 

There are also other changes to accommodate these updates. And an update/refactor to the run timing structs. These structs now record individual component run time and produce a summary of the 10 slowest components at the end of a model run. It also fixes the use of the timings in multi-model runs.


Batch21 (changes requested):
couple of comments but otherwise looks good

Batch21 (approved):

PR #431, opened 2025-08-12
feat: Add node components.

### LINKED ISSUE #362: Clarify the use of `node_indices_for_constraints`
This API is a bit ambiguous at the moment. I think we need to add some documentation and clarification on what this method should do. That will help developers writing new nodes.

For the near term I think it should:

1. Be used to provide the nodes which provide the flow for consideration in aggregated constraints and virtual storage nodes. 
2. Return an error when requested for storage type nodes. I do not think we currently have a valid use case for a storage node being used as a reference in either aggregated node or virtual storage. 

In the medium-long term figure out an API where the user can specify which sub components might be used. For example, in a complex compound node with internal abstraction and compensation nodes the user in one case might want the abstraction flow to apply (e.g. VirtualStorage) and in another the compensation flow (e.g. aggregated node control flow ratios). This will require definition at the point the node is referenced; not in the node definition itself.

jetuk:
Add a new `NodeComponent` enum that is used to determine which sub-nodes are used in flow constraints. Each schema node specifies a sub-set enum, similar to the attributes, which is then used to return different core node indices as required.

To allow this there are now `NodeAttrReference` and `NodeComponentReference` which are used in different places depending on the context. Typically the aggregated and virtual nodes will take component references now.

There is additional module level documentation on attributes and components, and improved docstrings for several schema nodes.

The attribute and component enums are made public, which may also be useful for tools to determine which variants are allowed.

Added some additional test models for aggregated storage and virtual storage. The aggregated storage tests ensure that these work correctly with all the main storage nodes.

Added new invalid model tests framework. This is to test models which have a valid schema, but should error when trying to "build" into a core model. Currently there is only one invalid model to ensure that a flow node can not be used in an `AggregatedStorageNode`.

Fixes #362


Batch21 (changes requested):
I might have missed it somewhere, but it would be nice to have a test/example that uses a non-default node component as a constraint. 

Otherwise this looks good!

Batch21 (approved):

PR #453, opened 2025-08-20
feat: Add new model and network level error types.

jetuk:
Improve the handling of errors during model building. This will provide more context information about which part and component of the build process failed.

In addition, it should pass the "cause" of the resulting exception as any underlying `PyErr` (e.g. from a custom parameter, or loading via pandas) if one exists.

### DIFF
--- a/pywr-core/src/models/mod.rs
+++ b/pywr-core/src/models/mod.rs
@@ -29,7 +29,10 @@ impl ModelDomain {
         Self { time, scenarios }
     }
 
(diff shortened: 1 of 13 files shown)


Batch21:
just been testing this with the pandas test and it gives me the underlying error:

image

Is there any way to format the python error better?

Batch21:
> I think part of the issue is the `.unwrap()` in the Rust CLI. It just uses debug to print the error. More explicit handling and printing of the error would be better.

ok. Probably something to look at in another PR.

PR #454, opened 2025-08-20
feat: Merge all virtual storage nodes into a single node type.

jetuk:
This reduces some code duplication for what is mostly just a different definition of the `reset` field.


Batch21 (changes requested):
A few comments but I like this change!

Batch21 (approved):

PR #464, opened 2025-08-26
chore: allow a tolerance when checking test model outputs

Batch21:


jetuk (changes requested):
This looks good, and is probably more robust than comparing the file contents. We just need to use the `float-cmp` crate for the comparisons. That might allow for some of the trailing decimals to be removed.

You might also be able to configure the tolerances for each test?

jetuk (approved):
Just needs the conflict resolving.

PR #467, opened 2025-08-28
feat: Implement microlp solver.

jetuk:
This is a Rust based solver that uses the simplex method and can handle integer variables. A new feature is added to enable this solver. The API requires rebuilding the LP every solve, which is probably not very efficient.

This solver also uses f64::INFINITY for identifing unbounded rows and columns. The current solvers uses f64::MAX. The solver builders are updated so that this value can be set on a per solver basis.

Added it to the CI tests.


Batch21 (approved):
LGTM!

PR #474, opened 2025-09-02
feat: Add seasonal virtual storage reset.

jetuk:
Adds an active flag to virtual storage nodes. When not active the storage bounds become unbounded, the node is not updated with flows from the model, and it is reset.

This means the node will be at reset volume outside the active period.

The active period is configured when selecting a Seasonal reset period in the schema.

Add two new test models.


pywr-schema/src/nodes/virtual_storage.rs
```diff
@@ -39,6 +39,15 @@ pub struct AnnualReset {
     pub month: u8,
 }
 
+#[derive(serde::Deserialize, serde::Serialize, Clone, Debug, JsonSchema, PywrVisitAll)]
+#[serde(deny_unknown_fields)]
+pub struct SeasonalReset {
```
Batch21, line 44: It would be good to have the docstring of the virtual storage update to mention this reset option

PR #481, opened 2025-09-05
refactor: Refactor and merge aggregation functions.

jetuk:
There were two f64 aggregation function enums in pywr-core. These are merged together and shared between parameters and metric sets. The refactor moves the aggregation enums to a common module in both core and schema.

This also moves to a tagged enum representation for the variants. This will allow more complex functions, that require user data, to be implemented later.


pywr-core/src/agg_funcs.rs
```diff
@@ -0,0 +1,389 @@
...
+    Max,
+    CountNonZero,
+    CountFunc { func: fn(f64) -> bool },
+}
+
+impl AggFuncF64 {
+    /// Calculate the aggregation of the given values.
+    pub fn calc_period_values(&self, values: &[PeriodValue<f64>]) -> Option<f64> {
+        match self {
+            Self::Sum => Some(values.iter().map(|v| v.value * v.duration.fractional_days()).sum()),
```
Batch21, line 19: Is there a reason why fractional days is not accounted for in the Product, Min and Max variants? I might be missing something here...

Batch21 (approved):

PR #494, opened 2025-09-15
feat: Implement custom Python aggregation functions.

jetuk:
Refactors some of the Python loading code (e.g. args and kwargs) to be shared between the aggregation functions and Python parameter


Batch21 (approved):
All looks good to me. Only suggestion would be to add something to the book on it but  that's probably better in a separate PR at some point as I don't think there is anything on aggregation yet.

PR #501, opened 2025-09-16
feat: Add structured model results.

jetuk:
This changes a few things about the recorder state and improves the Python extension to allow retrieving dataframes.

- Recorder state is now moved/consumed in the finalise method.
- Recorders can instead return a new object that implements `RecorderFinalResult`.
- This provides a more consistent lifetime for the the simulation state, and separate data for the returned results.
- The new result structs are added to the Python extension, and a method is provided to convert the data to Polars dataframes where possible.
- The MultiNetworkModel is also updated to return results keyed by the network names.


pywr-core/src/models/simple.rs
```diff
@@ -88,6 +88,32 @@ impl From<ModelRunError> for PyErr {
...
+impl ModelResult {
+    /// Get a reference to the results map.
+    pub fn network_result(&self) -> &NetworkResult {
+        &self.network_result
+    }
+}
+
+#[cfg(feature = "pyo3")]
+#[pymethods]
+impl ModelResult {
```
Batch21, line 109: a `__repr__` method, showing some basic information on scenarios/speed/etc..., would be useful, especially for notebooks

pywr-core/src/models/simple.rs
```diff
@@ -88,6 +88,32 @@ impl From<ModelRunError> for PyErr {
     }
 }
 
+/// The results of a model run.
+///
+/// Only recorders which produced a result will be present.
+#[cfg_attr(feature = "pyo3", pyclass)]
+#[derive(Clone)]
+pub struct ModelResult {
+    network_result: NetworkResult,
```
Batch21, line 97: could the timings stats also be included here?

Batch21 (commented):
Played around with this in a notebook and it works well. Just a few suggestions..

PR #504, opened 2025-09-17
fix: add virtual storage set_constraints method in schema

Batch21:


pywr-schema/src/nodes/virtual_storage.rs
```diff
@@ -251,28 +251,10 @@ impl VirtualStorageNode {
         Ok(indices)
     }
     pub fn add_to_model(&self, network: &mut pywr_core::network::Network, args: &LoadArgs) -> Result<(), SchemaError> {
```
jetuk, line 253: Does `args` become unused here? It should be removed.

jetuk (commented):
I wonder if the other variants in `add_to_model` should also not require `LoadArgs`?

pywr-schema/src/nodes/virtual_storage.rs
```diff
@@ -251,28 +251,10 @@ impl VirtualStorageNode {
...
-
         let node_idxs = self.node_indices_for_flow_constraints(network, args)?;
 
-        let mut builder = VirtualStorageBuilder::new(self.meta.name.as_str(), &node_idxs)
-            .initial_volume(self.initial_volume.into())
-            .min_volume(min_volume)
-            .max_volume(max_volume)
-            .cost(cost);
+        let mut builder =
+            VirtualStorageBuilder::new(self.meta.name.as_str(), &node_idxs).initial_volume(self.initial_volume.into());
```
jetuk, line 257: Should we remove the `cost`, `min_volume`, and `max_volume` from the builder?

jetuk (changes requested):
Looks good. I wonder if we should remove those methods from the builder to ensure they are not used in future?

jetuk (approved):

PR #507, opened 2025-09-17
fix: Make nodes ignore VS costs by default.

jetuk:
This means virtual storage costs are ignored by any attached nodes by default. This is a more sensible default behvaiour for users. Using the virtual storage costs must now be set explicitly.


Batch21 (approved):
LGTM!

PR #512, opened 2025-09-19
feat: Add multi-model to Python extension.

jetuk:
Refactor Python extension code to remove wrappers, and expose schema and model structs directly. Add MultiNetworkModel to the extension so these models can be run from Python.


Batch21 (approved):
All seems to work well when testing and I think all the feature gating makes sense!

PR #519, opened 2025-09-22
fix: update storage attribute of ControlCurvePiecewiseInterpolatedParameter

Batch21:
This update converts it from a storage node reference to a metric reference. Some logic is also added to the v1 to v2 conversion to handle to determine if a node is virtual or not.

### DIFF
--- a/pywr-schema/src/parameters/control_curves.rs
+++ b/pywr-schema/src/parameters/control_curves.rs
@@ -1,7 +1,7 @@
 #[cfg(feature = "core")]
 use crate::error::SchemaError;
 use crate::error::{ComponentConversionError, ConversionError};
-use crate::metric::{Metric, NodeAttrReference};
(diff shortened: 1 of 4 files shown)


pywr-schema/src/parameters/control_curves.rs
```diff
@@ -106,15 +106,22 @@ impl TryFromV1<ControlCurveInterpolatedParameterV1> for ControlCurveInterpolated
         };
 
         // v1 uses proportional volume for control curves
-        let storage_node = NodeAttrReference {
-            name: v1.storage_node,
-            attribute: Some(NodeAttribute::ProportionalVolume),
+        let storage_metric = if conversion_data.virtual_nodes.contains(&v1.storage_node) {
```
jetuk, line 109: Is this going to be dependent on whether the virtual node was converted first or not (e.g. inline parameters)?

jetuk (approved):

PR #530, opened 2025-09-24
feat: add method to NetworkResult that returns names of available outputs

Batch21:


pywr-core/src/network.rs
```diff
@@ -624,6 +624,11 @@ impl NetworkResult {
             .and_then(|r| r.aggregated_value().map_err(|e| e.into()))
     }
 
+    /// An iterator over the names of all available outputs.
```
jetuk, line 627: This needs adding to the `pywr,.pyi` file.

jetuk (changes requested):

jetuk (approved):

PR #532, opened 2025-09-25
fix: update timeseries_pandas model for recent schema changes

Batch21:


jetuk:
Can the tests at least check deserialisation of this model without that feature?

PR #546, opened 2025-10-02
feat: Allow the CBC solver to be used from Python.

jetuk:


Batch21 (commented):
The Python CLI solver option needs updating with a 'cbc' choice, otherwise looks good

PR #552, opened 2025-10-03
feat: Initial implementation of hourly time-steps

jetuk:
Includes DiurnalProfile parameter that returns a different value for every hour.

Change the `Timestep` to a tagged enum in the schema, and improve the docstrings. The hourly and daily timesteps are non-zero u64s.

Test existing timeseries produces 3-hourly output.


Batch21 (approved):
All looks good. Tested a few different setups and all worked as expected.

PR #564, opened 2025-10-08
fix: Swap "rows" and "cols" keys in table lookup definition.

### LINKED ISSUE #556: Tables schema have confusing rows and cols fields
The "rows" and "cols" fields are backward. I.e. a row-based data should specify columns for look-up, and a vice-versa.

jetuk:
When using row orientated data you should specify the number of columns to use as the index. And vice-versa for column orientated data.

The change also impacts the mean of rows and cols keys in the "Both" lookup type. The internal function has been modified to follow this updated naming convention.

Fixes #556.


Batch21 (approved):
Approving this as the changes look good. Separately the book chapter on tables needs to be updated to account for shifts in the row numbers.

PR #569, opened 2025-10-17
feat: Introduce NodeSlot enum for edges.

jetuk:
Following a similar pattern to `NodeComponent` and `NodeAttribute` this update adds a new enum `NodeSlot` to capture global slot types. It then also adds local slot enums for each node that requires slots. These are separated into input and output slots. This improves the typing / validation of slots.

It also adds some methods for iterating the available slots.


Batch21 (approved):
This looks good. My only suggestions is that it it could do with a bit more testing. I think its just few test models that do so atm.

PR #599, opened 2025-12-09
feat: Allow parameters to (optionally) return value in after.

jetuk:
This refactors the calculation of parameters to allow them to optionally return values from their "after" method. This can be used to return post-solve values (e.g. deficits). It allows parameters to return one value in "before" (e.g. demand required) and another outcome in "after" (e.g. yield or power). This allows the same data to be reused in the parameter struct (or object).

To do the above the "calc" method has been renamed to "before" for consistency. Also derived metrics are now redundant and have been removed.

The internal state for managing storage has also be refactored to store the max_volume alongside the current volume. This makes it easier to ensure the proportional volume is computed correctly and only updated after the solve (or if volume is set directly). I.e. it makes sure that the proportional volume does not change during "before" if max volume is a parameter value.

A new listing showing how to use this feature in a Python parameter is given.


pywr-schema/src/nodes/reservoir.rs
```diff
@@ -701,9 +695,7 @@ impl ReservoirNode {
                                 name: self.meta().name.clone(),
                                 sub_name: None,
                             })?;
-                        let dm = DerivedMetric::NodeProportionalVolume(idx);
-                        let derived_metric_idx = network.add_derived_metric(dm);
-                        MetricF64::DerivedMetric(derived_metric_idx)
+                        MetricF64::NodeVolume(idx)
```
Batch21: Should return `NodeProportionalVolume`?

pywr-core/src/state/mod.rs
```diff
@@ -884,31 +873,75 @@ impl State {
...
+
+    fn get_parameter_index_before(&self, idx: GeneralParameterIndex<u64>) -> Result<u64, StateError> {
+        self.parameters_before
+            .general
+            .get_index(*idx)
+            .ok_or(StateError::GeneralIndexParameterIndexNotFound(idx))
+    }
+
+    fn get_parameter_index_after(&self, idx: GeneralParameterIndex<u64>) -> Result<u64, StateError> {
+        self.parameters_before
```
Batch21: should be `self.parameters_after`?

pywr-core/src/parameters/mod.rs
```diff
@@ -1575,7 +1685,7 @@ impl ParameterCollection {
                         .get_simple_mut_multi_state(*idx)
                         .ok_or(ParameterCollectionSimpleCalculationError::MultiIndexNotFound(*idx))?;
 
-                    p.compute(
+                    p.before(
```
Batch21: should be `.after`?

Batch21 (commented):
A few potential bugs below but I think this is a good solution.

A more general point would be to add some examples/information of how to use the return_type attribute into the book. I don't think the agri example is currently referenced in it.

jetuk:
I have addressed the comments and also referenced the new agri listing in the book under the custom parameters section.

Batch21 (approved):

PR #617, opened 2026-02-09
chore: Bump polars to v0.53.

jetuk:
This also allows updating pyo3 to v0.27, ndarray to v0.17 and hdf5-metno to v0.12


Batch21 (approved):
Should this be rebase so the two commits can be retained?

PR #671, opened 2026-04-25
fix(core): Fix incorrect default lower bound.

jetuk:
Three of the solvers were using f64::MIN, the smallest positive non-zero number, instead of -f64::MAX the largest negative number. This is used as the lower bounds on unbounded rows.

In practice this would only impact inactive virtual nodes.


LaurenPetch1 (approved):
Looks good as far as I can tell and seems to make sense

PR #713, opened 2026-06-30
feat(core): Implement thread pool for multi-model solves.

jetuk:
Multi-model now respects the parallel and threads settings.

Removes some TODOs.


Batch21 (approved):
looks good

PR #714, opened 2026-07-01
refactor: Use a const for float equality tolerance

jetuk:


Batch21 (approved):
LGTM. Only query is if the plan is to use `FLOAT_EQ_TOLERANCE` across the project, is thresholds.rs the best place for it top live?

jetuk:
That's a fair comment. I think there's only one other similar use of this constant in an interpolate function. There are other uses for 1e-6, but a lot are default tolerances for asserts. I can refactor this.

Batch21 (approved):

PR #724, opened 2026-07-15
Enable optimisation of x values in RbfProfileParameter

s-simoncelli:
In the current implementation only the `y` values can be optimised. The new implementation allows passing all the variable values


s-simoncelli:
I'll close this as I just realised `VariableParameter` is also implemented for `u32`

PR #744, opened 2026-08-05
feat: Move to jiff datetime library.

jetuk:
Chrono is now soft-deprecated. Much of the Rust ecosystem is switching to jiff. These changes move to Jiff for all but the interface with polars. Polars is still using chrono, and we cannot drop the chrono dependency until polars does.

There are some minor integer type changes. The biggest change is the removal of the `Date` enum in favour of just using `jiff::civil::DateTime` directly. However, this changes the roundtrip deserlisation/serialisation behaviour losing the information on whether the user provided a simple date or a date-time.

The conversion tests currently fail because they are not updated to reflect this change.


jetuk:
I have put  a separate commit on here to bump `pywr-v1-schema` to the latest version which also does the jiff migration.

pywr-core/src/timestep.rs
```diff
@@ -330,44 +338,28 @@ impl TimeDomainBuilder {
...
-            .enumerate()
-            .map(|(i, dates)| {
-                let duration = dates[1] - dates[0];
-                Timestep::new(dates[0], i, duration.into())
-            })
-            .collect::<Vec<Timestep>>();
+        let mut timesteps: Vec<Timestep> = Vec::new();
+        let mut current = self.start;
+        while current <= self.end {
+            let next = current + span;
```
Batch21: I think next needs to be

```rust
let next = self.start + span * (timesteps.len() as i64 + 1);
```

If the timestep is a monthly it will currently go 02-28 -> 03-28 instead of 02-28 -> 03-31

pywr-core/src/timestep.rs
```diff
@@ -330,44 +338,28 @@ impl TimeDomainBuilder {
         let duration = polars::time::Duration::parse(frequency);
```
Batch21, line 319: should we check here that the duration is not zero or negative? Previously this would have be done by the polars range function.

pywr-schema/src/nodes/virtual_nodes/virtual_storage.rs
```diff
@@ -74,23 +74,17 @@ impl TryInto<pywr_core::virtual_storage::VirtualStorageReset> for VirtualStorage
...
-            VirtualStorageReset::Annual(annual) => {
-                let reset_month = annual.month.try_into()?;
-                pywr_core::virtual_storage::VirtualStorageReset::DayOfYear {
-                    day: annual.day as u32,
-                    month: reset_month,
-                }
-            }
+            VirtualStorageReset::Annual(annual) => pywr_core::virtual_storage::VirtualStorageReset::DayOfYear {
+                day: annual.day,
+                month: annual.month,
```
Batch21, line 79: Now that the core month values are `i8` not chrono `Months` should we have some validation here that the values are in the correct range? Same for the other month fields below

pywr-core/src/timestep.rs
```diff
@@ -15,68 +13,74 @@ const MILLISECS_IN_HOUR: i64 = 1000 * 60 * 60;
...
 }
 
-impl PartialEq<TimeDelta> for PywrDuration {
-    fn eq(&self, other: &TimeDelta) -> bool {
-        self.0 == *other
+impl PartialEq<Span> for PywrDuration {
+    fn eq(&self, other: &Span) -> bool {
+        self.0
+            .compare(SpanCompare::from(other).days_are_24_hours())
+            .expect("Span comparison failed!")
```
Batch21: are the expects ok in these methods? one way to avoid them would be to switch to using `SignedDuration` instead of `Span` as the innner type but I don't know if that removes some required functinality?

Batch21 (changes requested):
A few queries but runs ok for me.

pywr-core/src/timestep.rs
jetuk: Is this any different to `polars::time::date_range`?

```
>>> pl.date_range(date(2000, 2, 28), date(2000, 8, 31), '1mo', eager=True).alias("date")
shape: (7,)
Series: 'date' [date]
[
        2000-02-28
        2000-03-28
        2000-04-28
        2000-05-28
        2000-06-28
        2000-07-28
        2000-08-28
]
```

pywr-core/src/timestep.rs
jetuk: If we wrap `SignedDuration` in `PywrDuration` then we lose the knowledge it was "1 month" or "2 days". It just becomes a (possibly) different duration for each timestep. This is possibly OK as mainly just need to know the length of each one anyway. However, it means that each timestep is no longer equal. Currently we have a `PartialEq` implementation that is used to validate each timestep is equal in the series. I am not sure that is really required though.

pywr-core/src/timestep.rs
Batch21: I think the issue is with 29th–31st start days, where the date will get clamped to the 28th for all months after the first Feb.

Batch21 (approved):
changes look good

PR #745, opened 2026-08-10
feat: Allow specifying calculation phase in some parameters.

jetuk:
This folows on from #726 to do the following:
- pywr-core: Internally allocate space only for before and after parameters that are registered to supply values in those phases. Two new index types are added for this.
- pywr-core: Validate consumer phasing with parameter return values during network build. This should reject invalid combinations at build time and avoid spurious missing values at runtime.
- pywr-schema: Allow specifying the phase for Aggregated and AggregatedIndex parameters in the schema.
- pywr-schema: Correctly construct parameter references for appropriate phase when building metric sets from all parameters.
- pywr-schema: Add tests models for the issue identified in #725 (i.e. calculating and saving total deficit).

What's left to do is implement alternative phase (i.e. after) for some of the other parameter types.


pywr-core/src/parameters/aggregated_index.rs
```diff
@@ -152,7 +164,7 @@ impl ParameterBuilder<u64> for AggregatedIndexParameterBuilder {
         self: Box<Self>,
         resolution_maps: &ResolutionMaps,
     ) -> Result<MaybeBuiltParameter<u64>, ParameterBuildError> {
-        let metrics = resolve_metric_u64_vec!(self, &self.metrics, resolution_maps, "metrics");
+        let metrics = resolve_metric_u64_vec!(self, &self.metrics, resolution_maps, self.phase, "metrics");
 
         let meta = self.meta;
         let agg_func = self.agg_func;
```
Batch21, line 180: does the builder need to match on `phase` when processing the metrics like in `AggregatedParameterBuilder`?

pywr-core/src/parameters/mod.rs
```diff
@@ -3437,194 +3577,18 @@ mod tests {
...
-        assert_eq!(
-            state
-                .get_general_parameter_value(hook_index, ParameterReturnValue::After,)
-                .unwrap(),
-            0.0
-        );
-    }
-
-    #[test]
-    fn simple_before_parameter_can_depend_on_simple_before_value() {
```
Batch21, line 3468: do these removed tests need replacing or updating?

pywr-core/src/metric.rs
```diff
@@ -760,6 +780,442 @@ impl From<f64> for UnresolvedMetricF64 {
     }
 }
 
+/// Resolve a [`ParameterIndex<f64>`] to a [`MetricF64`] using the provided [`ParameterReturnValue`]
+/// and [`MetricConsumerPhase`]. This function is used to determine if a parameter can be resolved to a metric
+/// based on the phase in which the consumer is using the metric and the return value of the parameter.
+///
+/// If the parameter cannot be resolved to a metric, an error is returned.
+fn resolve_parameter_index_f64_to_metric_f64(
+    name: &ParameterName,
```
Batch21, line 798: Should we have some unit tests that check that the correct errors are produced for different combinations of return values and phases.

Also, would it be possible to combined some of the logic from the `resolve_parameter_index_*` functions into a share function?

pywr-schema/src/parameters/mod.rs
```diff
@@ -178,6 +184,47 @@ impl Parameter {
...
+            Self::WeeklyProfile(_) => ParameterPhase::Before,
+            Self::UniformDrawdownProfile(_) => ParameterPhase::Before,
+            Self::Max(_) => ParameterPhase::Before,
+            Self::Min(_) => ParameterPhase::Before,
+            Self::MultiThreshold(_) => ParameterPhase::Before,
+            Self::Negative(_) => ParameterPhase::Before,
+            Self::Polynomial1D(_) => ParameterPhase::Before,
+            Self::Threshold(_) => ParameterPhase::Before,
+            Self::TablesArray(_) => ParameterPhase::Before,
+            Self::Python(_) => ParameterPhase::Before,
```
Batch21, line 212: The Python parameter could be After or Both which could cause an error when this is used by by metrics sets. Do these phases need to be determined dynamically from the registered parameters instead?

pywr-core/src/metric.rs
```diff
@@ -760,6 +780,442 @@ impl From<f64> for UnresolvedMetricF64 {
...
+    idx: ParameterIndex<f64>,
+    parameter_return_value: ParameterReturnValue,
+    consumer_phase: MetricConsumerPhase,
+) -> Result<MetricF64, MetricF64ResolutionError> {
+    match idx {
+        // Constant and simple can always be resolved to a metric, regardless of the consumer phase
+        // as long as the parameter return value is "before".
+        ParameterIndex::Const(idx) => match parameter_return_value {
```
Batch21, line 806: The other `resolve_parameter_index_*` function do not match on `parameter_return_value` from constant paramters like this one does.

pywr-core/src/parameters/delay.rs
```diff
@@ -253,7 +253,10 @@ impl ParameterBuilder<f64> for DelayParameterBuilder<UnresolvedMetricF64, f64> {
         self: Box<Self>,
         resolution_maps: &ResolutionMaps,
     ) -> Result<MaybeBuiltParameter<f64>, ParameterBuildError> {
-        let metric = resolve_metric_f64!(self, self.metric, resolution_maps, "metric");
+        // Phase is hardcoded to "both" for this parameter, as it only implements the
+        // `GeneralBeforeParameter` and `GeneralAfterParameterHook` traits.
+        let phase = MetricConsumerPhase::Both;
```
Batch21: shouldn't this be `After`. The delay parameter only accesses the metric value in that phase.

pywr-core/src/parameters/rolling.rs
```diff
@@ -320,7 +320,10 @@ impl ParameterBuilder<f64> for RollingParameterBuilder<UnresolvedMetricF64, f64,
         self: Box<Self>,
         resolution_maps: &ResolutionMaps,
     ) -> Result<MaybeBuiltParameter<f64>, ParameterBuildError> {
-        let metric = resolve_metric_f64!(self, self.metric, resolution_maps, "metric");
+        // Phase is hardcoded to "both" for this parameter, as it only implements the
+        // `GeneralBeforeParameter` and `GeneralAfterParameterHook` traits.
+        let phase = MetricConsumerPhase::Both;
```
Batch21: Similarly as for `DelayParameterBuilder` shouldn't this be `After`?

pywr-core/src/parameters/hydropower.rs
```diff
@@ -191,12 +191,27 @@ impl ParameterBuilder<f64> for HydropowerTargetParameterBuilder {
...
         resolution_maps: &ResolutionMaps,
     ) -> Result<MaybeBuiltParameter<f64>, ParameterBuildError> {
-        let actual_flow = resolve_optional_metric_f64!(self, &self.actual_flow, resolution_maps, "actual_flow");
-        let target = resolve_optional_metric_f64!(self, &self.target, resolution_maps, "target");
-        let max_flow = resolve_optional_metric_f64!(self, &self.max_flow, resolution_maps, "max_flow");
-        let min_flow = resolve_optional_metric_f64!(self, &self.min_flow, resolution_maps, "min_flow");
+        // Determine which calculation phase(s) the parameter will be used in
+        // based on whether the target and actual flow are provided.
+        // If neither is provided, return an error.
+        let phase = match (self.target.is_some(), self.actual_flow.is_some()) {
```
Batch21, line 197: won't the phase here vary across the metrics? some will be access in before and some in after?

pywr-core/src/metric.rs
```diff
@@ -760,6 +780,442 @@ impl From<f64> for UnresolvedMetricF64 {
...
+                        }),
+                    }
+                }
+                (ParameterReturnValue::AfterOrElseInitial, MetricConsumerPhase::Both) => {
+                    // The consumer is using the metric in both "before" and "after" phases, and the
+                    // parameter is providing an "after" value. However, they have specified that using any
+                    // initial value is acceptable in the "before" phase, so we can resolve it to a metric provided the
+                    // parameter index contains an "after" index.
+                    match idx.after {
+                        Some(after_idx) => Ok(MetricF64::ParameterAfterMulti {
```
Batch21, line 1250: This could result in a runtime error as it will try and look up a value in an empty hashmap?

pywr-core/src/parameters/asymmetric.rs
```diff
@@ -71,14 +71,17 @@ pub struct AsymmetricSwitchIndexParameterBuilder {
     meta: ParameterMeta,
     on_parameter: UnresolvedMetricU64,
     off_parameter: UnresolvedMetricU64,
+    phase: MetricConsumerPhase,
 }
 
 impl AsymmetricSwitchIndexParameterBuilder {
-    pub fn new(name: ParameterName, on_parameter: UnresolvedMetricU64, off_parameter: UnresolvedMetricU64) -> Self {
+    /// Create a new builder for [`AsymmetricSwitchIndexParameter`] that is evaluated in the "before" phase.
+    pub fn before(name: ParameterName, on_parameter: UnresolvedMetricU64, off_parameter: UnresolvedMetricU64) -> Self {
```
Batch21, line 79: Some parameters, like this one, define a new before constructor method while others (like the max parameter) hardcode a before phase variable in their existing build methods. Is there a reasons for this? Should it be made consistent?

pywr-schema/tests/deficit-agg3.json
```diff
@@ -0,0 +1,133 @@
+{
```
Batch21, line 1: This model isn't tested anywhere?

pywr-core/src/parameters/aggregated.rs
```diff
@@ -119,19 +119,43 @@ impl ConstParameter<f64> for AggregatedParameter<ConstantMetricF64> {
...
 pub struct AggregatedParameterBuilder {
     meta: ParameterMeta,
     agg_func: AggFuncF64,
     metrics: Vec<UnresolvedMetricF64>,
+    phase: MetricConsumerPhase,
 }
 
 impl AggregatedParameterBuilder {
-    pub fn new(name: ParameterName, agg_func: AggFuncF64) -> Self {
+    /// Create a new builder for [`AggregatedParameter`] that is evaluated in the "after" phase.
```
Batch21: "before"

Batch21 (changes requested):
I've gone through this in some detail and think the approach is good. A few queries/suggestions below.

pywr-core/src/parameters/asymmetric.rs
jetuk: I think this was just a bit of WIP. I was going to do this as the main framework, and then go back through and implement it for other parameters. So, I renamed this one but it doesn't support `after` at the moment.

pywr-core/src/parameters/hydropower.rs
jetuk: Isn't that what the match statement is doing? Working out which phases this parameter will be in?

pywr-core/src/metric.rs
jetuk: Yes, that's true. I think a follow-up idea would allow parameters to provide initial values.

pywr-core/src/parameters/aggregated_index.rs
jetuk: Done.

pywr-core/src/parameters/mod.rs
jetuk: I have added more unit tests.

pywr-schema/src/parameters/mod.rs
jetuk: Currently the parameter (in core) is hardcoded to be "before". I'll fix this along with other parameters in a follow-up.

pywr-core/src/metric.rs
jetuk: Fixed.

pywr-core/src/parameters/delay.rs
jetuk: Fixed.

pywr-core/src/parameters/rolling.rs
jetuk: Fixed.

pywr-core/src/parameters/aggregated.rs
jetuk: Fixed.

pywr-core/src/parameters/hydropower.rs
```diff
@@ -191,12 +191,55 @@ impl ParameterBuilder<f64> for HydropowerTargetParameterBuilder {
...
+        // Actual flow and target are used to determine the phases calculated.
+        let actual_flow = resolve_optional_metric_f64!(self, &self.actual_flow, resolution_maps, phase, "actual_flow");
+        let target = resolve_optional_metric_f64!(self, &self.target, resolution_maps, phase, "target");
+        // Water elevation is used in both phases; so can be safely resolved if provided.
         let water_elevation =
-            resolve_optional_metric_f64!(self, &self.water_elevation, resolution_maps, "water_elevation");
+            resolve_optional_metric_f64!(self, &self.water_elevation, resolution_maps, phase, "water_elevation");
+        // min and max flow have a chance of being unused if they are supplied but "before" is not computed.
+        let (min_flow, max_flow) = match phase {
+            MetricConsumerPhase::Before => {
```
Batch21: shouldn't this be the `After` variant? min and max flow are accessed in before.

Batch21 (approved):
Updates look good

PR #750, opened 2026-08-16
fix(schema): Visit virtual nodes when visiting network metrics

Batch21:
This also adds tests to ensure that the vistor can reach metrics in all possible locations within a netwok schema.


jetuk (approved):

jetuk:
LGTM

PR #751, opened 2026-08-16
feat(schema): Enforce unique node names within a network

Batch21:
Because node names in pywr-core are based on both the name and sub-name attrs and because schema nodes can add multiple nodes to core it is possible for two schema nodes with the same name to be added to core without raising an error. This can result in a misleading error such as `UnresolvedNode`. This PR solves this by checking for duplicates in the schema, though there might be a better way of doing this.

### DIFF
--- a/pywr-schema/src/network.rs
+++ b/pywr-schema/src/network.rs
@@ -5,7 +5,9 @@ use crate::ConversionError;
 use crate::data_tables::DataTable;
 #[cfg(feature = "core")]
 use crate::data_tables::{LoadedTableCollection, TableCollectionLoadError};
-use crate::error::{ComponentConversionError, DuplicateNodeName, SchemaError};
+#[cfg(feature = "core")]
+use crate::error::SchemaError;
+use crate::error::{ComponentConversionError, DuplicateNodeName, ValidationError};
 use crate::metric::Metric;
 use crate::metric_sets::MetricSet;
 #[cfg(feature = "core")]
@@ -47,7 +49,7 @@ pub enum NetworkSchemaBuildError {
     #[error("Network schema validation failed: {source}")]
     Validation {
         #[source]
-        source: Box<SchemaError>,
+        source: ValidationError,
     },
     #[error("Circular node reference(s) found.")]
     CircularNodeReference,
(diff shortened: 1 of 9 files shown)


pywr-schema/src/network.rs
```diff
@@ -445,6 +449,61 @@ impl NetworkSchema {
...
+    ///
+    /// Every entry of [`NetworkSchema::nodes`] and [`NetworkSchema::virtual_nodes`] must have a
+    /// unique name.
+    ///
+    /// Duplicates must be rejected here because name resolution in this crate is
+    /// first-match-wins (see [`NetworkSchema::get_node_by_name`]). A duplicate therefore does
+    /// not fail on its own: it silently binds every reference to whichever entry appears first
+    /// and leaves the other unreachable.
+    ///
+    /// `pywr-core` cannot catch this for us. Its duplicate check runs over the *expanded* core
```
jetuk: The only question to ask ourselves is whether this should be caught in pywr-core during build. It would be possible to use a namespace builder (or something) to permit the construction of only one parent node name. 

I guess is it more or less complex to do it there. Is the addition of a validation step going to confusing or difficult to maintain.

pywr-schema/src/network.rs
```diff
@@ -445,6 +449,61 @@ impl NetworkSchema {
...
+    /// nodes, keyed by name **and** sub-name. One schema node may expand to several core nodes,
+    /// and composite types such as [`crate::nodes::DelayNode`] emit only sub-named ones.
+    pub fn validate(&self) -> Result<(), SchemaError> {
```
jetuk: I might argue if we do this that we should separate the error types between `validation` and `add_to_network` ValidationError.

pywr-schema/src/network.rs
```diff
@@ -445,6 +449,61 @@ impl NetworkSchema {
...
+    /// Validation is **not** exhaustive. It checks that the schema is internally unambiguous,
+    /// not that it can be built, which should be done using [`NetworkSchema::add_to_network`]
+    /// instead.
+    ///
+    /// # Unique node names
+    ///
+    /// Every entry of [`NetworkSchema::nodes`] and [`NetworkSchema::virtual_nodes`] must have a
+    /// unique name.
+    ///
+    /// Duplicates must be rejected here because name resolution in this crate is
```
jetuk: This reads a bit like internal documentation, not documentation of the function itself.

jetuk (commented):
The changes look good to me. I would just like to discuss whether we should add this to the core builder infrastructure in some fashion.

pywr-schema/src/network.rs
Batch21: I've made a separate error type.

pywr-schema/src/network.rs
Batch21: I've simplified this

jetuk (approved):

PR #758, opened 2026-08-18
feat: Slot validation API

jetuk:
Add a new API for validating slots against a node's schema.

Also correct the logic in the availability of the Spill slot
in the `ReservoirNode`. It is only available if "spill" is
set to the `Spill` variant.

### DIFF
--- a/pywr-schema/src/nodes/reservoir.rs
+++ b/pywr-schema/src/nodes/reservoir.rs
@@ -110,6 +110,7 @@ node_component_subset_enum! {
     }
 }
 
+#[derive(PartialEq)]
 pub enum ReservoirOutputNodeSlot {
     Storage,
     Compensation,
@@ -240,6 +241,7 @@ pub struct ReservoirNode {
 
 impl ReservoirNode {
     pub const DEFAULT_COMPONENT: ReservoirNodeComponent = ReservoirNodeComponent::Compensation;
+    const DEFAULT_OUTPUT_SLOT: ReservoirOutputNodeSlot = ReservoirOutputNodeSlot::Storage;
 
     /// Get the node's metadata.
     pub(crate) fn meta(&self) -> &NodeMeta {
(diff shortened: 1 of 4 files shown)


pywr-schema/src/nodes/reservoir.rs
```diff
@@ -251,13 +253,33 @@ impl ReservoirNode {
...
 
-    pub fn iter_output_slots(&self) -> impl Iterator<Item = NodeSlot> + '_ {
-        [
-            ReservoirOutputNodeSlot::Storage.into(),
-            ReservoirOutputNodeSlot::Compensation.into(),
-            ReservoirOutputNodeSlot::Spill.into(),
-        ]
-        .into_iter()
+    pub fn iter_output_slots(&self) -> impl Iterator<Item = ReservoirOutputNodeSlot> + '_ {
+        let mut slots = vec![ReservoirOutputNodeSlot::Storage, ReservoirOutputNodeSlot::Compensation];
```
Batch21: The `Compensation` slot should be conditional on if a compensation flow has been set?

pywr-schema/src/nodes/river_split_with_gauge.rs
```diff
@@ -102,16 +102,35 @@ pub struct RiverSplitWithGaugeNode {
...
             .chain(self.splits.iter().enumerate().map(|(i, split)| match &split.slot_name {
-                Some(name) => NodeSlot::User { name: name.clone() },
-                None => NodeSlot::Split { position: i },
+                Some(name) => RiverSplitWithGaugeOutputNodeSlot::User { name: name.clone() },
+                None => RiverSplitWithGaugeOutputNodeSlot::Split { position: i },
             }))
     }
 
+    /// Validate the output slot and return the corresponding `ReservoirOutputNodeSlot`.
+    pub fn output_slot(&self, slot: Option<&NodeSlot>) -> Result<RiverSplitWithGaugeOutputNodeSlot, SchemaError> {
```
Batch21, line 117: Is this stricter than `output_connecters` in that it will reject Split values being provided for slot with a user name? I'd say that this method is probabaly correct and `output_connecters` needs fixing.

Batch21 (changes requested):
couple of small queries

Batch21 (approved):

PR #759, opened 2026-08-19
feat(schema): Add a visitor for node references in a schema

Batch21:


jetuk (changes requested):
These changes look good to me. The only comment is that the node visitor is not implemented for the multi-network schema.

Batch21:
> These changes look good to me. The only comment is that the node visitor is not implemented for the multi-network schema.

I don't think the existing visitors have implementations for multi-models. Is that something to implement here or would a separate PR to add multi-model visitor implementations be better?

jetuk:
Fair point. Let's merge this and add that as an issue.

jetuk (approved):

PR #762, opened 2026-08-19
feat: Implement merge for NetworkSchema.

jetuk:
Initial commit of a basic merge function. It modifies a base network schema in-place by consuming another one. Merge errors on duplicates unless they are a placeholder type. Placeholders are replaced in-place.

Metric sets are treated different with the metric entries merged if the names match.

Added additional placeholder types for tables, outputs and timeseries.

Added additional API for finding other components on the schema.

### DIFF
--- a/pywr-schema/src/data_tables/mod.rs
+++ b/pywr-schema/src/data_tables/mod.rs
@@ -144,7 +144,7 @@ impl CsvDataTable {
     }
 }
 
-/// An external table of data that can be referenced
+/// A placeholder for an external table of data that can be referenced
 #[derive(serde::Deserialize, serde::Serialize, Debug, Clone, JsonSchema)]
 pub struct PlaceholderTable {
     pub meta: TableMeta,
--- a/pywr-schema/src/data_tables/mod.rs
+++ b/pywr-schema/src/data_tables/mod.rs
@@ -57,6 +57,7 @@ pub struct TableMeta {
 #[strum_discriminants(name(DataTableType))]
 pub enum DataTable {
     CSV(CsvDataTable),
+    Placeholder(PlaceholderTable),
 }
 
 impl DataTable {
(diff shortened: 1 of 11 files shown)


pywr-schema/src/network.rs
```diff
@@ -606,6 +719,177 @@ impl NetworkSchema {
...
+                    Some(existing_node) => {
+                        if existing_node.is_placeholder() {
+                            *existing_node = v_node;
+                        } else {
+                            return Err(NetworkMergeError::DuplicateNodeName(v_node.name().to_string()));
+                        }
+                    }
+                    None => {
+                        // Check if the node name exists in the virtual nodes list
+                        if self.get_virtual_node_index_by_name(v_node.name()).is_some() {
```
Batch21: should this call `get_node_index_by_name` instead?

pywr-schema/src/network.rs
```diff
@@ -606,6 +719,177 @@ impl NetworkSchema {
...
+
+    /// Merge another [`NetworkSchema`] into this one.
+    ///
+    /// This will combine the nodes, virtual nodes, edges, and parameters of both networks.
+    /// If there are any duplicate node or parameter names, an error will be returned. However,
+    /// placeholder types (e.g. [`crate::nodes::PlaceholderNode`]) are replaced.
+    ///
+    /// Metric sets are merged by name, with the metrics of any metric sets with the same name being
+    /// combined. Other information in the metric set (e.g. filters) is **not** merged.
+    pub fn merge(&mut self, other: NetworkSchema) -> Result<(), NetworkMergeError> {
```
Batch21, line 781: Is it worth having a comment in the docstring to say an error might leave `self` partially merged? An alternative would be to have two phases in the method; one to check that it is possible then one to apply it.

pywr-schema/src/network.rs
```diff
@@ -606,6 +719,177 @@ impl NetworkSchema {
...
+        }
+
+        // Merge metric sets. There are no placeholder metric sets. Instead, we merge the metrics
+        // of any metric sets with the same name.
+        if let Some(other_metric_sets) = other.metric_sets {
+            for ms in other_metric_sets {
+                match self.get_metric_set_by_name_mut(&ms.name) {
+                    Some(existing_ms) => {
+                        // Merge the metrics of the existing metric set with the new one.
+                        if let Some(existing_metrics) = &mut existing_ms.metrics {
```
Batch21, line 899: should this error if the `aggregator` or `filters` attributes don't match.

pywr-schema/src/data_tables/mod.rs
```diff
@@ -135,6 +144,12 @@ impl CsvDataTable {
     }
 }
 
+/// An external table of data that can be referenced
```
Batch21: docstring should say its a placeholder?

Batch21 (changes requested):
A few small queries/suggestions:

pywr-schema/src/network.rs
jetuk: I have added the comment to the docstring.

A 2-stage process might work, but unless we use unsafe in the second stage you'll still have to handle some errors there (even if they are unexpected).

pywr-schema/src/network.rs
jetuk: I thought you might just want a set with a list of metrics in the `other` network, and that get merged with the original set. Otherwise, if you change the original set options you'll need to the other one too.

Batch21 (approved):

PR #765, opened 2026-08-21
feat: add phase options to max/min parameters

LaurenPetch1:
Includes:
* MaxParameter
* MinParameter
* NegativeParameter
* NegativeMaxParameter
* NegativeMinParameter

### DIFF
--- a/pywr-book/src/developers-guide/adding-a-parameter.md
+++ b/pywr-book/src/developers-guide/adding-a-parameter.md
@@ -45,7 +45,7 @@ To allow the parameter to be used in the model a "builder" is required. This str
 `ParameterBuilder<T>` trait. This builder will be used by the schema to create the parameter when it is loaded from a model
 file. Builders will typically have the same fields as the parameter itself, but will use "unresolved" types
 (e.g. `UnresolvedMetricF64` or `UnresolvedNode`). This allows the builder to be created without first resolving the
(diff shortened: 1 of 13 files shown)


jetuk (commented):
Changes look good to me. However, can you check whether the text in the book surrounding the listing needs updating to reflect these changes?

pywr-book/src/developers-guide/adding-a-parameter.md
```diff
@@ -42,12 +42,17 @@ The `ParameterMeta` struct is used to store the metadata for all parameters and
 ```
 
 To allow the parameter to be used in the model a "builder" is required. This struct must implement the
-`ParameterBuilder<T>` trait. This builder be used by the schema to create the parameter when it is loaded from a model
+`ParameterBuilder<T>` trait. This builder will be used by the schema to create the parameter when it is loaded from a model
 file. Builders will typically have the same fields as the parameter itself, but will use "unresolved" types
 (e.g. `UnresolvedMetricF64` or `UnresolvedNode`). This allows the builder to be created without first resolving the
-dependencies of the parameter. The `build` function is then used to resolve the dependencies and create the parameter
+dependencies of the parameter. he `build` function is then used to resolve the dependencies and create the parameter
```
jetuk: Typo here

jetuk (approved):

PR #766, opened 2026-08-24
feat: Support metrics providing a value aligned to the current phase.

jetuk:
The current implementation, while allowing metrics to compute values in each phase, does not allow referencing a parameter's value aligned to the current phase. This change allows fetching the parameter's value as per the current the calculation phase. Therefore, it allows computing more complex chains of "Both" values. For example, calculating aggregations of both values with a single aggregation parameter.

### DIFF
--- a/pywr-book/src/custom_parameters.md
+++ b/pywr-book/src/custom_parameters.md
@@ -69,7 +69,7 @@ from pywr import ParameterInfo
 
 
 def current_timestep(
-        info: ParameterInfo, a: float, b: float, some_condition: str = "foo"
+    info: ParameterInfo, a: float, b: float, some_condition: str = "foo"
 ) -> float:
     """Return the current time step."""
     match some_condition:
--- a/pywr-book/src/custom_parameters.md
+++ b/pywr-book/src/custom_parameters.md
@@ -1,22 +1,19 @@
 # Extending functionality with custom parameters
 
-Parameters are a core part of Pywr, allowing you to define how your model behaves.
-While Pywr comes with a wide range of built-in parameters, you may find that you need to create custom parameters to
-suit your specific modelling needs.
-This guide will walk you through the process of creating custom parameters in Pywr.
+Parameters are a core part of Pywr, allowing you to define how your model behaves. While Pywr comes with a wide range of
+built-in parameters, you may find that you need to create custom parameters to suit your specific modelling needs. This
+guide will walk you through the process of creating custom parameters in Pywr.
 
-Currently, Pywr supports custom parameters that are defined in Python.
-If your parameter is general enough, you may want to consider contributing it to the Pywr project.
-If you do, please see the [Developers Guide](./developers-guide/adding-a-parameter.md) for more information on how to do
-this.
+Currently, Pywr supports custom parameters that are defined in Python. If your parameter is general enough, you may want
+to consider contributing it to the Pywr project. If you do, please see
+the [Developers Guide](./developers-guide/adding-a-parameter.md) for more information on how to do this.
 
 ## Python functions
 
-The simplest way to create a custom parameter is to define a Python function.
-This function should accept at least one argument, which is a `ParameterInfo` object.
-This object contains information from the model, such as the current time step, scenario index,
-and any metric values that have been requested.
-Additional arguments can also be passed to the function.
+The simplest way to create a custom parameter is to define a Python function. This function should accept at least one
+argument, which is a `ParameterInfo` object. This object contains information from the model, such as the current time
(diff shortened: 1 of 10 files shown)


pywr-core/src/metric.rs
```diff
@@ -1547,15 +1763,15 @@ impl From<u64> for UnresolvedMetricU64 {
...
 ///
 /// If the parameter cannot be resolved to a metric, an error is returned.
 fn resolve_parameter_index_u64_to_metric_u64(
     name: &ParameterName,
     idx: ParameterIndex<u64>,
-    parameter_return_value: ParameterReturnValue,
+    parameter_return_value: UnresolvedParameterReturnValue,
     consumer_phase: MetricConsumerPhase,
 ) -> Result<MetricU64, MetricU64ResolutionError> {
     match idx {
```
Batch21, line 1308: The other resolve functions match on `parameter_return_value` for const and simple indexes here.

pywr-core/src/metric.rs
```diff
@@ -150,16 +160,14 @@ pub enum MetricF64 {
...
     },
-    ParameterBeforeF64(GeneralBeforeValueIndex<f64>),
-    ParameterAfterF64(GeneralAfterValueIndex<f64>),
-    ParameterBeforeU64(GeneralBeforeValueIndex<u64>),
-    ParameterAfterU64(GeneralAfterValueIndex<u64>),
-    ParameterBeforeMulti {
-        index: GeneralBeforeValueIndex<MultiValue>,
-        key: String,
+    ParameterF64 {
+        indices: ParameterReturnValue<f64>,
```
Batch21: would this attribute be better renamed to something like `source` as `indices` is already used in this enum for something else? The renamed could also be applied to `Metricu64`.

pywr-core/src/metric.rs
```diff
@@ -109,6 +109,16 @@ impl SimpleMetricF64 {
     }
 }
 
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum ParameterReturnValue<T> {
```
Batch21: It maybe a little confusing that this shares a name with a schema type. Could be renamed something like `ParameterValueSource`?

pywr-schema/tests/hydropower1.json
```diff
@@ -0,0 +1,174 @@
...
+        "turbine_elevation": 100.0,
+        "min_head": 0.0,
+        "efficiency": 1.0,
+        "water_density": 1000.0,
+        "flow_unit_conversion": 1.0,
+        "energy_unit_conversion": 1e-6
+      },
+      {
+        "meta": {
+          "name": "total_flow_target"
```
Batch21: this and the `total_power` param could be removed from this model as they are not tested?

Batch21 (commented):
A few small suggestions below.

My main comment is whether the growth in the size of the `resolve_parameter_*` functions now justifies creating a share function that captures the matching on `return_value` and `consumer_phase`. This might make adding some unit test for this logic easier.

pywr-book/src/custom_parameters.md
```diff
@@ -344,19 +335,19 @@ calculations after the resource allocation has been completed. Typically, this i
...
+computes a crop yield in `after` based on the allocated water and the water requirement.
 
 [//]: # (@formatter:off)
 
 ```python,ignore
 {{ #include ../py-listings/agri-parameter/agri_parameter.py}}
 ```
 
-When referring to the parameter in the model, the `return_value` field can be used to specify whether to use the value 
+When referring to the parameter in the model, the `source` field can be used to specify whether to use the value
```
Batch21: This should still be `return_value`

pywr-book/src/custom_parameters.md
```diff
@@ -344,19 +335,19 @@ calculations after the resource allocation has been completed. Typically, this i
...
 
 ```python,ignore
 {{ #include ../py-listings/agri-parameter/agri_parameter.py}}
 ```
 
-When referring to the parameter in the model, the `return_value` field can be used to specify whether to use the value 
+When referring to the parameter in the model, the `source` field can be used to specify whether to use the value 
 returned by the `before` or `after` method. By default, the value returned by the `before` method is used, but if you 
-want to use the value from the `after` method you can set `return_value` to `"After"`. The example below shows how to 
+want to use the value from the `after` method you can set `source` to `"After"`. The example below shows how to
```
Batch21: should also still be `return_value` instead of `source`

Batch21 (commented):
couple of queries on changes to the docs otherwise I think this looks good to merge.

Batch21 (approved):

PR #776, opened 2026-08-27
feat: add parameter phases to the control curve parameters

LaurenPetch1:


pywr-core/src/parameters/control_curves/interpolated.rs
```diff
@@ -104,12 +173,10 @@ impl ParameterBuilder<f64> for ControlCurveInterpolatedParameterBuilder {
         self: Box<Self>,
         resolution_maps: &ResolutionMaps,
     ) -> Result<MaybeBuiltParameter<f64>, ParameterBuildError> {
-        // Phase is hardcoded to "before" for this parameter, as it only implements the `GeneralBeforeParameter` trait.
-        let phase = MetricConsumerPhase::Before;
-        let metric = resolve_metric_f64!(self, self.metric, resolution_maps, phase, "metric");
+        let metric = resolve_metric_f64!(self, self.metric, resolution_maps, self.phase, "metric");
         let control_curves =
```
jetuk, line 177: Could we add the validation of the length of `control_curves` and `values` to this method? I think that would be a useful addition to error early with a helpful message.

pywr-core/src/parameters/control_curves/piecewise.rs
```diff
@@ -41,33 +41,57 @@ impl GeneralBeforeParameter<f64> for PiecewiseInterpolatedParameter {
...
+}
+
+
+fn piecewise_interpolated(x: f64, control_curves: &[f64], values: &[[f64; 2]], maximum: f64, minimum: f64) -> Result<f64, GeneralCalculationError> {
+    let mut cc_previous_value = maximum;
+    for (idx, &control_curve) in control_curves.iter().enumerate() {
+        if x >= control_curve {
+            let v = values
+                .get(idx)
+                .ok_or_else(|| GeneralCalculationError::OutOfBoundsError {
```
jetuk: This error handling is different to the interpolated one where `[]` is used which would panic if wrong. I wonder if we can do the same here by validating the number of values and control curves in the builder.

pywr-core/src/parameters/control_curves/simple.rs
```diff
@@ -107,14 +155,12 @@ impl ParameterBuilder<f64> for ControlCurveParameterBuilder {
         self: Box<Self>,
         resolution_maps: &ResolutionMaps,
     ) -> Result<MaybeBuiltParameter<f64>, ParameterBuildError> {
-        // Phase is hardcoded to "before" for this parameter, as it only implements the `GeneralBeforeParameter` trait.
-        let phase = MetricConsumerPhase::Before;
-        let metric = resolve_metric_f64!(self, self.metric, resolution_maps, phase, "metric");
+        let metric = resolve_metric_f64!(self, self.metric, resolution_maps, self.phase, "metric");
 
         let control_curves =
```
jetuk, line 160: Again would be good to validate the lengths here.

pywr-schema/src/parameters/control_curves.rs
```diff
@@ -375,13 +399,28 @@ impl ControlCurvePiecewiseInterpolatedParameter {
...
-            self.maximum.unwrap_or(1.0),
-            self.minimum.unwrap_or(0.0),
-        );
+        let name = ParameterName::new(&self.meta.name, parent);
+
+        let mut builder = match self.phase {
+            ParameterPhase::Before => {
+                pywr_core::parameters::PiecewiseInterpolatedParameterBuilder::before(name,
+                                                                                     metric,
+                                                                                     self.maximum.unwrap_or(1.0),
```
jetuk: a local `maximum` and `minimum` variable would make this a little less verbose by removing the repeated `.unwrap_or(..)`

jetuk (changes requested):
This looks good to me. It would be good to add the validation of number of control curves and values to make sure everything fails early. If that can't be done now then let's make an issue so it is not forgotten.

LaurenPetch1:
@jetuk I've added checks on the length of values relative to control curves but does it need another check to confirm that control_curves isn't empty or would that have already errored if it was?

jetuk:
> @jetuk I've added checks on the length of values relative to control curves but does it need another check to confirm that control_curves isn't empty or would that have already errored if it was?

If the builder would produce a parameter that errors because there's no control curves then yes. If the parameter calculation does work with zero control curves, then I think it is odd, but probably OK to allow. We should view these changes to pywr-core as though someone was using it independently (i.e. not with prior knowledge of what pywr-schema may have already checked).

pywr-core/src/parameters/control_curves/interpolated.rs
```diff
@@ -226,4 +226,16 @@ mod tests {
         assert_eq!(control_curve_interpolated(0.1, &control_curves, &values), 45.0);
 
     }
+
+    #[test]
+    fn test_control_curve_empty_interpolated() {
+        let control_curves = vec![];
+        let values = vec![10.0, 20.0];
```
jetuk, line 233: I think this needs a test if `values` is empty or length 1.

pywr-core/src/parameters/control_curves/piecewise.rs
```diff
@@ -218,6 +218,18 @@ mod test {
         assert_eq!(piecewise_interpolated(0.25, &control_curves, &values, maximum, minimum).unwrap(), -55.0);
     }
 
+    #[test]
+    fn test_piecewise_interpolated_empty() {
+        let control_curves = vec![];
+        let values = vec![[10.0, 0.0]];
```
jetuk, line 224: Empty values should be a build error

pywr-core/src/parameters/control_curves/simple.rs
```diff
@@ -210,4 +210,13 @@ mod tests {
         assert_eq!(calculate_control_curve(0.2, &control_curves, &values).unwrap(), 30.0);
         assert_eq!(calculate_control_curve(0.1, &control_curves, &values).unwrap(), 50.0);
     }
+
+    #[test]
+    fn test_calculate_control_curves_empty() {
+        let control_curves = vec![];
+        let values = vec![10.0, ];
```
jetuk, line 217: Test for empty.

jetuk (changes requested):
Tests look good for control curves. However, I think the values need build errors if there are not enough of them.

LaurenPetch1:
@jetuk I think this should be ok now on the basis that if values is empty it would be picked up by the build errors comparing length of control curves and values

PR #785, opened 2026-09-02
feat: Refactor and improve Python extension

jetuk:
Several things here are tidied up. This is working towards
being able to build pywr-core with the wasm32 target.

- Removed polars dependency from pywr-core. This should
speed-up pywr-core only builds. It was used for some
duration parsing (that is now ported to jiff), and a trait
providing a dataframe. The trait is refactored to provide
an iterable of a simple Rust type. Conversion to Dataframe
is now done in pywr-python.
- ahash is also only required when pyo3 feature is enabled.
- Moved most of the Python classes to wrapper structs in
pywr-python. This is instead of wrapping the "raw" structs
in pywr-core.
- Added Pywr specific exceptions to the Python extension
and mapped various error types to those exceptions.
- Fixed some errors in .pyi stub file and added the
new exceptions.

Outstanding changes:

- Similar tasks of moving pyclass derives out of pywr-schema. This
is more complex due to the wrapping of the Metric enum. That might
need a more complex design in the extension.
- rand is only needed for the test-utils module in pywr-core. This
module could be feature gated for tests only and rand made a
dev-dependency.
- Feature gating polars might be a useful thing to speed-up compile
times.


pywr-core/src/timestep.rs
```diff
@@ -314,28 +241,7 @@ impl TimeDomainBuilder {
...
-            duration.months().months()
-        } else {
-            Span::new()
-                .months(duration.months())
-                .weeks(duration.weeks())
-                .days(duration.days())
-                .nanoseconds(duration.nanoseconds())
-        };
-
+    fn generate_timesteps_from_span(&self, span: Span) -> Result<Vec<Timestep>, TimeDomainBuilderError> {
```
Batch21, line 244: I think this function should check  for a negative or zero span like the previous function did for the duration.

pywr-python/Cargo.toml
```diff
@@ -27,9 +28,11 @@ ipm-simd = ["pywr-core/ipm-simd", "pywr-schema/ipm-simd"]
 [dependencies]
 # The Github workflow is configured to use the minimum version corresponding to the `abi3-pyXX` feature enabled here.
 # Please remember to update the workflow if changing the ABI version.
-pyo3 = { workspace = true, features = ["extension-module", "macros", "chrono", "abi3-py310"] }
+jiff = { workspace = true }
+pyo3 = { workspace = true, features = ["extension-module", "macros", "jiff-02", "abi3-py312"] }
```
Batch21: do we need to bump the `requires-python` in `pyproject.toml` to >=3.12

pywr-core/src/parameters/py.rs
```diff
@@ -235,7 +356,7 @@ impl PyClassParameter {
                 {
                     let mut info_mut = info_bind.borrow_mut();
                     info_mut.timestep = *ctx.timestep;
-                    info_mut.scenario_index = ctx.scenario_index.clone();
+                    info_mut.scenario_index = Arc::new(ctx.scenario_index.clone());
```
Batch21, line 359: does the scenario_index need updating here? it should be static through the run?

Batch21 (changes requested):
a few small comment otherwise looks good

pywr-python/Cargo.toml
jetuk: Actually I bumped this to use `#[pyclass(extends=PyException)]` which requires >=3.12. However, I then ended up not using  that. So, it could be reverted back to 3.10.

pywr-core/src/parameters/py.rs
jetuk: Perhaps not, but it is what it did before.

jetuk: I do wonder if updating this object is a premature optimisation, and it might be cleaner to create a new instance each time.

Batch21 (approved):

PR #789, opened 2026-09-06
refactor: Use log and env_logger crates instead of tracing.

jetuk:
Both log and env_logger are more approriate for Pywr. Also pyo3-log supports log and not tracing. This simplifies the dependencies and is generally more appropriate for Pywr. The CLI now uses env_logger to display logs from the Pywr crates.


Batch21 (approved):
LGTM

PR #792, opened 2026-09-06
refactor: Refactor ScenarioDomainBuilder

### LINKED ISSUE #784: Rationalise ModelSchemaBuildError
Scenario building should be redesigned to remove the need for `ModelSchemaBuildError::ScenarioBuilderError` variant.

https://github.com/pywr/pywr-next/blob/f382dc6582ac6cf1ef616e8090d60b0457cd5019/pywr-schema/src/model.rs#L382

`ModelSchemaBuildError::CoreModelDomainError` is also unused.

jetuk:
Refactor `ScenarioDomainBuilder` to accept `ScenarioGroupBuilder` instead of pre-built `ScenarioGroup`s. This allows the top-level build method to return all errors and allows simplification of the errors that schema needs to handle.

The builders are also refactor to take and return mutable references instead of moving self. This aligns with the more recent builders implemented in the rest of pywr-core.

Fixes #784.

### DIFF
--- a/pywr-core/src/scenario.rs
+++ b/pywr-core/src/scenario.rs
@@ -103,6 +103,7 @@ impl ScenarioGroup {
 }
 
 /// A builder for defining a subset of scenarios to run for a scenario group.
+#[derive(Clone, Debug)]
 enum ScenarioGroupSubsetBuilder {
     Slice { start: usize, end: usize },
     Indices(Vec<usize>),
@@ -114,6 +115,7 @@ enum ScenarioGroupSubsetBuilder {
 /// This allows for defining the subset of scenarios to run for a group in different ways
 /// (e.g. as a slice, as specific indices, or as labels). It also allows for defining labels for the
 /// group, which can be used in the subset definition.
+#[derive(Clone, Debug)]
 pub struct ScenarioGroupBuilder {
     name: String,
     size: usize,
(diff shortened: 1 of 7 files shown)


pywr-core/src/scenario.rs
```diff
@@ -280,24 +278,17 @@ impl ScenarioDomainBuilder {
...
     ///
     /// - [`ScenarioDomainBuilderError::DuplicateGroupName`] if a group with the same name already exists in the builder
-    pub fn with_group(mut self, group: ScenarioGroup) -> Result<Self, ScenarioDomainBuilderError> {
-        for g in &self.groups {
-            if g.name == group.name {
-                return Err(ScenarioDomainBuilderError::DuplicateGroupName(group.name.clone()));
-            }
-        }
-
+    pub fn with_group(&mut self, group: ScenarioGroupBuilder) -> &mut Self {
```
Batch21, line 280: The docstring needs updating for the new type.

Batch21 (changes requested):
One minor comment but otherwise I think this is good to merge

pywr-core/src/scenario.rs
jetuk: Done

Batch21 (approved):

PR #796, opened 2026-09-09
feat(schema): Replace VisitNodeReferences with an owner-aware VisitReferences

Batch21:


pywr-schema/src/network.rs
```diff
@@ -275,53 +275,131 @@ impl VisitMetrics for NetworkSchema {
...
+
+        for output in self.outputs.as_deref_mut().into_iter().flatten() {
+            output.visit_references_mut(visitor);
         }
     }
 }
 
 impl NetworkSchema {
+    /// Visit every reference together with the element holding it.
+    pub fn visit_owned_references<F: FnMut(Owner<'_>, Reference<'_>)>(&self, visitor: &mut F) {
```
jetuk, line 334: These owners might not be the director "holder" of the reference. It could be nested somewhere. Is that a problem?

pywr-schema/src/network.rs
```diff
@@ -275,53 +275,131 @@ impl VisitMetrics for NetworkSchema {
...
+        }
+
+        for parameter in self.parameters.as_deref_mut().into_iter().flatten() {
+            let owner_name = parameter.name().to_string();
+            parameter.visit_references_mut(&mut |reference| visitor(Owner::Parameter(&owner_name), reference));
+        }
+
+        for metric_set in self.metric_sets.as_deref_mut().into_iter().flatten() {
+            let owner_name = metric_set.name.clone();
+            metric_set
```
jetuk: Why is the trait not implemented for metric sets directly?

jetuk (commented):
Just a couple of questions.

pywr-schema/src/network.rs
Batch21: It should have been. Fixed!

pywr-schema/src/network.rs
Batch21: I think its ok. Its meant to just given the top level group that the reference is in.

jetuk (approved):

PR #799, opened 2026-09-10
feat(schema): Validate edges in NetworkSchema::validate

Batch21:


pywr-schema/src/error.rs
```diff
@@ -32,6 +33,33 @@ impl std::fmt::Display for DuplicateNodeName {
     }
 }
 
+/// The reason an [`Edge`] is invalid.
+#[derive(Error, Debug, Clone, PartialEq, Eq)]
+pub enum EdgeProblem {
```
jetuk, line 38: The error messages here could be formatted as complete sentences. I think that follows the style of the other errors.

pywr-schema/src/error.rs
```diff
@@ -32,6 +33,33 @@ impl std::fmt::Display for DuplicateNodeName {
...
+    #[error("there is no node named `{0}`")]
+    UnknownNode(String),
```
jetuk, line 40: The other variants here distinguish between "from" and "to". Should we have `UnknownFromNode` and `UnknownToNode` as different variants?

pywr-schema/src/error.rs
```diff
@@ -50,6 +78,12 @@ pub enum ValidationError {
     /// A timestep frequency string that parses, but is zero or negative.
     #[error("The timestep frequency `{freq}` is not a positive duration.")]
     NonPositiveFrequency { freq: String },
+    /// One or more edges could not connect the nodes they name.
+    #[error(
```
jetuk, line 82: This has the possibility of create a very long error message. Should we avoid that by either limiting the length or just stating the number of errors?

The same issue is true for the `DuplicateNodeNames` variant above.

pywr-schema/src/network.rs
```diff
@@ -623,11 +623,67 @@ impl NetworkSchema {
...
+    ///
+    /// The following conditions are checked, with the first problem found being
+    /// returned:
+    ///
+    /// - Both ends name an entry of `nodes`; a virtual node is not an edge end.
+    /// - The two ends are different nodes.
+    /// - Each slot is one that the node at that end has.
+    /// - The `from_node` can provide flow, and the `to_node` can receive it.
+    ///
+    /// All but the second are checks `pywr-core` makes only while building. The second is
```
jetuk, line 636: Is this a bug in pywr-core?

pywr-schema/src/network.rs
```diff
@@ -623,11 +623,67 @@ impl NetworkSchema {
...
     /// Validate the network schema.
     ///
-    /// This checks that the schema is unambiguous, not that it can be built; use
-    /// [`NetworkSchema::add_to_network`] for the latter. See [`ValidationError`] for the
-    /// problems that are detected.
+    /// This checks that the schema is unambiguous and that its edges could be made, not that the
+    /// whole model can be built; use [`NetworkSchema::add_to_network`] for the latter. See
+    /// [`ValidationError`] for the problems that are detected, and
+    /// [`NetworkSchema::validate_edge`] for the edge rules in particular.
     pub fn validate(&self) -> Result<(), ValidationError> {
```
jetuk, line 687: I wonder if this should return a `Vec<ValidationProblem>` that captures the duplicate node and edge problems, the other variants are model level validation issues.

pywr-schema/src/error.rs
```diff
@@ -32,6 +33,33 @@ impl std::fmt::Display for DuplicateNodeName {
...
+pub struct EdgeValidationError {
+    /// The invalid edge.
+    pub edge: Edge,
+    /// The first problem [`crate::NetworkSchema::validate_edge`] found with it.
+    pub problem: EdgeProblem,
+}
+
 /// A problem found by [`crate::ModelSchema::validate`] or [`crate::NetworkSchema::validate`].
 #[derive(Error, Debug, Clone, PartialEq, Eq)]
 pub enum ValidationError {
```
jetuk, line 65: See other comment. It might make sense to separate the two network level errors into their own tuple.

pywr-schema/src/network.rs
```diff
@@ -1053,6 +1101,199 @@ mod tests {
...
+                    node_type: NodeType::Reservoir,
+                    slot: NodeSlot::Spill,
+                }
+            )]
+        );
+    }
+
+    /// A duplicate name makes an edge's ends ambiguous, so the names are reported first.
+    #[test]
+    fn test_validate_reports_duplicate_names_before_edges() {
```
jetuk, line 1272: Do you think this would be better if it told you both problems? See other comments.

jetuk (changes requested):
The changes look correct to me. However, I wonder if the reporting function should gather all network problems together. And whether these should be separated from the model problems. Such that at he end there is a full list of all problems in a model or network.

pywr-schema/src/network.rs
Batch21: I don't think it is. Core doesn't track which of its nodes come from which composite nodes so can't easily enforce this. Feels more like a schema-side task to me as we probably want to avoid core having to worry about the configuration of composite nodes?

pywr-schema/src/error.rs
Batch21: fixed

pywr-schema/src/error.rs
Batch21: updated

Batch21:
@jetuk I've updated this to return all validation problems and addressed your other comments.

pywr-schema/src/error.rs
```diff
@@ -26,21 +27,47 @@ impl std::fmt::Display for DuplicateNodeName {
...
-    )]
-    DuplicateNodeNames(Vec<DuplicateNodeName>),
+pub enum EdgeProblem {
+    /// The `from_node` is not an entry of `nodes`.
+    #[error("There is no node named `{0}` to connect from.")]
+    UnknownFromNode(String),
+    /// The `to_node` is not an entry of `nodes`.
+    #[error("There is no node named `{0}` to connect to.")]
+    UnknownToNode(String),
+    #[error("A node cannot be connected to itself.")]
```
jetuk: Could this include the node name?

pywr-schema/src/error.rs
```diff
@@ -52,6 +79,111 @@ pub enum ValidationError {
...
+}
+
+impl std::fmt::Display for NetworkValidationError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let subject = match &self.name {
+            Some(name) => format!("The network `{name}`"),
+            None => "The network".to_string(),
+        };
+
+        write_problems(
```
jetuk: What's the guidance on having a multiline Display implementation? The alternative is a method.. like `write_pretty(...)`

jetuk (changes requested):
One suggestion and a question.

pywr-schema/src/error.rs
Batch21: added

pywr-schema/src/error.rs
Batch21: seems like its good practice to keep display to 1 line. I've done this and added report methods that print the full list of problems.

pywr-schema/src/error.rs
```diff
@@ -26,21 +27,47 @@ impl std::fmt::Display for DuplicateNodeName {
...
-    )]
-    DuplicateNodeNames(Vec<DuplicateNodeName>),
+pub enum EdgeProblem {
+    /// The `from_node` is not an entry of `nodes`.
+    #[error("There is no node named `{0}` to connect from.")]
+    UnknownFromNode(String),
+    /// The `to_node` is not an entry of `nodes`.
+    #[error("There is no node named `{0}` to connect to.")]
+    UnknownToNode(String),
+    #[error("The `{0}` node cannot be connected to itself.")]
```
jetuk: I think this should be: "The node `{0}` cannot be ..."

Also, I think the ones below should follow this pattern: "A `{node_type}` node ..."

jetuk (changes requested):
One grammatical change for the error messages.

pywr-schema/src/error.rs
Batch21: for the ones below I've gone with `This `{node_type}` node...` as the slots availble for some nodes depend on how its setup (e.g the Reservoir node)

jetuk: Should it therefore also include the node name?

Batch21: Done. also added more info to the other error messages.

jetuk (approved):

PR #800, opened 2026-09-10
feat(core): Write PYWR_VERSION and PYWR_VERSION_STR attrs to H5 outputs

Batch21:


pywr-core/src/recorders/hdf.rs
```diff
@@ -296,6 +296,27 @@ fn write_pywr_metadata(file: &hdf5_metno::File) -> Result<(), Hdf5Error> {
             source,
         })?;
 
+    const VERSION_MAJOR: &str = env!("CARGO_PKG_VERSION_MAJOR");
+    let major_version: i64 = VERSION_MAJOR
```
jetuk: I think you can make this integer const with new Rust as well.

Batch21: Done!

pywr-core/src/recorders/hdf.rs
```diff
@@ -296,6 +296,27 @@ fn write_pywr_metadata(file: &hdf5_metno::File) -> Result<(), Hdf5Error> {
             source,
         })?;
 
+    const VERSION_MAJOR: i64 = match i64::from_str_radix(env!("CARGO_PKG_VERSION_MAJOR"), 10) {
+        Ok(v) => v,
+        Err(_) => panic!("CARGO_PKG_VERSION_MAJOR cannot be parsed as an integer"),
+    };
+
+    let attr = root
+        .new_attr::<i64>()
```
jetuk, line 305: Is this the same type that Pywr v1.x creates? I would guess a u8 would probably be better for a version number.

jetuk (approved):

pywr-core/src/recorders/hdf.rs
Batch21: As far as I can tell, an `i64` matches the pywr v1 type, which gets converted to a numpy int before before being serialised.

PR #801, opened 2026-09-13
feat: Replace polars with arrow-rs for internal timeseries handling. 

jetuk:
Replace polars for handling time-series internally.

This removes the automatic resampling of time-series data to align
with the model time domain. That is now the responsibility of the
user.

The use of arrow requires CSV data to be in ISO8601 format. Hence
the change to the test data. This is probably a more sensible
default to avoid issues with day-first date formats.

For Rust only data loading arrow CSV and IPC formats are supported.
Polars and Pandas support is retained via use of Python callbacks
similar to the existing Pandas support. A generic Python format
is also supported to allow users to supply their own loading
function.

Parquet format is added using the parquet crate.

Compile times are significantly improve by the removal of polars
on the Rust side.

Finally, an initial draft of a book chapter on time-series
data providers is added.


pywr-core/src/parameters/array.rs
```diff
@@ -292,19 +332,42 @@ impl ParameterBuilder<f64> for Array2ParameterBuilder<f64> {
...
         if let Some(subset) = resolution_maps
             .domain
             .scenarios()
             .group_scenario_subset(&self.scenario_group)?
         {
-            array = subset_array2(&array, subset)?;
+            array = array
+                .into_iter()
+                .enumerate()
+                .filter_map(|(i, a)| subset.contains(&i).then_some(a))
```
Batch21: This could give the wong columns as the subset index is its position in the subset list not in the scenario group. same for the u64 code.

pywr-core/src/parameters/array.rs
```diff
@@ -88,125 +82,158 @@ impl SimpleParameter<u64> for Array1Parameter<u64> {
...
     fn name(&self) -> &ParameterName {
         &self.meta.name
     }
     fn build(
         self: Box<Self>,
-        _resolution_maps: &ResolutionMaps,
+        resolution_maps: &ResolutionMaps,
     ) -> Result<MaybeBuiltParameter<f64>, ParameterBuildError> {
+        let mut array =
+            cast(&self.array, &DataType::Float64).map_err(|source| ParameterBuildError::ArrayCastError {
```
Batch21: looks like `cast` might silently set values it can't parse to null. there is a `cast_with_options` version that can be set to error is these situations.

pywr-schema/src/time_series/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+    ) -> Result<Self, Self::Error> {
+        let meta: ParameterMeta = v1.meta.try_into_v2(parent_node, conversion_data)?;
+        let mut ts_name = meta.name.clone();
+
+        if let Some(url) = v1.url {
+            // If there is a URL then this entry must be converted into a time series
+            let mut pandas_kwargs = v1.pandas_kwargs;
+
+            let time_col = match pandas_kwargs.remove("index_col") {
+                Some(v) => v.as_str().map(|s| s.to_string()),
```
Batch21: it looks like a index_col with a 0 value will set time_col to None here. this is quite a common value so maybe worth accounting for.

pywr-schema/src/time_series/polars.rs
```diff
@@ -0,0 +1,94 @@
...
+
+            Ok(lt)
+        }
+
+        /// Make a copy of the kwargs and add the default value for try_parse_dates if not already present.
+        fn make_kwargs(&self) -> HashMap<String, serde_json::Value> {
+            let mut kwargs = self.kwargs.clone().unwrap_or_default();
+
+            if !kwargs.contains_key("try_parse_dates") {
+                kwargs.insert("try_parse_dates".to_string(), serde_json::Value::Bool(true));
```
Batch21: `try_parse_dates` is not a valid args for the function that read json and parquet

pywr-schema/src/time_series/parquet_ts.rs
```diff
@@ -0,0 +1,97 @@
+use crate::parameters::ParameterMeta;
+use crate::{Checksum, VisitPaths};
+use pywr_schema_macros::skip_serializing_none;
+use schemars::JsonSchema;
+use std::path::{Path, PathBuf};
+
+/// A dataset that can be loaded using Apache Arrow.
+///
+/// This dataset is loaded using Apache Arrow. This is done using the Rust Arrow library to load
+/// the dataset.
```
Batch21, line 10: docstring needs updating?

pywr-schema/src/time_series/py.rs
```diff
@@ -0,0 +1,97 @@
...
+                self.time_col.as_deref(),
+                &None,
+                &kwargs,
+            )?;
+
+            Ok(lt)
+        }
+
+        /// Make a copy of the kwargs and add the default value for try_parse_dates if not already present.
+        fn make_kwargs(&self) -> HashMap<String, serde_json::Value> {
```
Batch21: update docstring

pywr-core/src/parameters/array.rs
```diff
@@ -88,125 +82,158 @@ impl SimpleParameter<u64> for Array1Parameter<u64> {
     }
 }
 
+/// Builder for `Float64ArrayParameter` or `UInt64ArrayParameter`.
```
Batch21: update docstring types

pywr-schema/src/time_series/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+    TimeSeriesError { name: String, source: TimeSeriesError },
+    #[error("A time series with name '{0}' already exists.")]
+    DuplicateTimeSeriesName(String),
+    #[error("TimeSeries '{0}' not found in collection.")]
+    TimeSeriesNotFound(String),
+    #[error(
+        "The time series dataset '{0}' has more than one column of data so a column or scenario name must be provided for any reference"
+    )]
+    TimeSeriesColumnOrScenarioRequired(String),
+    #[error("The time series dataset is empty and has no columns of data.")]
```
Batch21: this should include the name

pywr-schema/src/time_series/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+            .get(name)
+            .ok_or(LoadedTimeSeriesCollectionError::TimeSeriesNotFound(name.to_string()))?;
+
+        let (array, time_array) = match df.num_columns() {
+            0 => {
+                return Err(LoadedTimeSeriesCollectionError::TimeSeriesHasNoColumns(
+                    name.to_string(),
+                ));
+            }
+            1 => (df.column(0), None),
```
Batch21, line 426: should this error with time col not found?

Cargo.toml
```diff
@@ -60,7 +61,7 @@ nalgebra = "0.35"
 nalgebra-sparse = "0.12"
 num = "0.4"
 ocl = "0.19"
-polars = { version = "0.55", default-features = false }
+parquet = { version = "59.3", features = ["arrow"] }
 pyo3 = { version = "0.29" }
 pyo3-log = "0.13"
 pyo3-polars = { version = "0.28", default-features = false }
```
Batch21: can this be removed?

Batch21 (changes requested):
A few things that look like they need fixing but this is a good change. A couple more suggestions:

- PywrDuration::duration_string can be deleted.
- polars could be an optional dependency in the pyproject.toml.

pywr-core/src/parameters/array.rs
jetuk: Was the original function incorrect as well?

pywr-schema/src/time_series/mod.rs
jetuk: This is just using the data as a raw array? Or do you think that should not be allowed?

pywr-core/src/parameters/array.rs
```diff
@@ -325,59 +401,226 @@ impl ParameterBuilder<u64> for Array2ParameterBuilder<u64> {
...
+
         // If there is a scenario subset then we can reduce the data to align with the scenarios
         // that are actually used in the model.
         if let Some(subset) = resolution_maps
             .domain
             .scenarios()
             .group_scenario_subset(&self.scenario_group)?
         {
-            array = subset_array2(&array, subset)?;
+            array = subset.iter().filter_map(|&index| array.get(index).cloned()).collect();
```
Batch21: This is going to silently drop out of range indexes which result in a array that is not the correct shape. This could be captured here with a nicer error message, rather than less clear one later on when the array is first used.

pywr-python/tests/models/time-series-formats/.~lock.inflow.csv#
```diff
@@ -0,0 +1 @@
+,james,fedora,17.09.2026 14:01,file:///home/james/.config/libreoffice/4;
```
Batch21: File needs removing

pywr-core/src/parameters/array.rs
```diff
@@ -88,125 +82,171 @@ impl SimpleParameter<u64> for Array1Parameter<u64> {
...
+                to: DataType::Float64,
+                source,
+            }
+        })?;
+
+        if let Some(time_array) = &self.time_array {
+            // Convert to DateTime array and align with the time-steps in the model.
+            array = subslice_data_for_time_domain(&array, time_array, resolution_maps.domain.time())?;
+        }
+
```
Batch21, line 160: It looks like the cast above won't catch values that are already null in data (e.g. a empty cell) and the null-ness of these values is then ignored when they are read(https://docs.rs/arrow-array/latest/arrow_array/array/struct.PrimitiveArray.html#method.value)

All simple null check could be added here:

```rust
f array.null_count() > 0 {
    return Err(/* new ArrayContainsNulls error */);
}
```

pywr-core/src/parameters/array.rs
```diff
@@ -88,125 +82,171 @@ impl SimpleParameter<u64> for Array1Parameter<u64> {
...
+                to: DataType::UInt64,
+                source,
+            }
+        })?;
+
+        if let Some(time_array) = &self.time_array {
+            // Convert to DateTime array and align with the time-steps in the model.
+            array = subslice_data_for_time_domain(&array, time_array, resolution_maps.domain.time())?;
+        }
+
```
Batch21, line 203: null check also needed here

pywr-core/src/parameters/array.rs
```diff
@@ -292,19 +345,42 @@ impl ParameterBuilder<f64> for Array2ParameterBuilder<f64> {
...
+        let f64_array = array
+            .into_iter()
+            .map(|a| {
+                cast_with_options(&a, &DataType::Float64, &options)
+                    .map_err(|source| ParameterBuildError::ArrayCastError {
+                        from: a.data_type().clone(),
+                        to: DataType::Float64,
+                        source,
+                    })
+                    .map(|a| a.as_primitive::<Float64Type>().clone())
```
Batch21: I think a null check is also required inside this map before returning.

pywr-book/src/external-data/timeseries.md
```diff
@@ -0,0 +1,182 @@
...
+  "kwargs": {"sheet_name": "inflow"}
+}
+```
+
+### Polars
+
+`"type": "Polars"` also uses the Python environment. It requires a Python-enabled Pywr build,
+and the Python packages `polars` and `pyarrow`. The built-in loader supports `.csv`, `.parquet`,
+and `.json` files. Its `kwargs` are forwarded to the selected Polars reader. Pywr supplies
+`"try_parse_dates": true` unless it is explicitly provided.
```
Batch21: This is no longer true

pywr-schema/src/time_series/py.rs
```diff
@@ -0,0 +1,91 @@
...
+pub struct PythonTimeSeries {
+    pub meta: ParameterMeta,
+    /// The Python module where the function is defined.
+    pub module: String,
+    /// The name of the function to call within the Python module.
+    pub function: String,
+    pub time_col: Option<String>,
+    /// Path to the dataset. If this is a relative path, it will be resolved relative to the provided data path.
+    pub path: PathBuf,
+    /// Keyword arguments to pass to the relevant Pandas load function.
```
Batch21: this comment is stale

pywr-python/tests/test_models.py
```diff
@@ -1,8 +1,10 @@
+import json
 from pathlib import Path
 
 import h5py
 import numpy as np
 import pandas
+import polars
 import polars as pl
```
Batch21, line 7: polars is imported twice

pywr-schema/src/time_series/mod.rs
```diff
@@ -0,0 +1,817 @@
...
+use crate::error::ComponentConversionError;
+use crate::parameters::ParameterMeta;
+use crate::v1::{ConversionData, TryFromV1, TryIntoV2};
+use crate::visit::{Reference, ReferenceMut, VisitPaths, VisitReferences};
+#[cfg(feature = "core")]
+use arrow::{
+    array::{Array, ArrayRef},
+    record_batch::RecordBatch,
+};
+use arrow_ts::ArrowTimeSeries;
```
Batch21: should be made pub? and also `ArrowFormat` and `ParquetTimeSeries`

Batch21 (changes requested):

pywr-core/src/parameters/array.rs
```diff
@@ -292,19 +353,55 @@ impl ParameterBuilder<f64> for Array2ParameterBuilder<f64> {
...
         let mut array = self.array;
 
+        if let Some(time_array) = &self.time_array {
+            // Convert to DateTime array and align with the time-steps in the model.
+            array = subslice_datas_for_time_domain(&array, time_array, resolution_maps.domain.time())?;
+        }
+
         // If there is a scenario subset then we can reduce the data to align with the scenarios
         // that are actually used in the model.
         if let Some(subset) = resolution_maps
```
Batch21, line 363: If there is not a subset, the array shape is not checked against the scenario size. Does this need a `if array.len() != scenario_group_size {` check prior to the subset check?

pywr-core/src/parameters/array.rs
```diff
@@ -325,59 +422,237 @@ impl ParameterBuilder<u64> for Array2ParameterBuilder<u64> {
...
```
Batch21, line 432: Also needs the same check here?

Batch21 (changes requested):
sorry, couple more suggestions for what I think is a pre-existing issue.

Batch21 (approved):

PR #807, opened 2026-09-15
feat: Support after_hook method in Python class parameters.

jetuk:
This allows users to defined `after_hook` instead of `after` if they just want to update state and not return a value.

Also removes an unneeded generic type on the GeneralAfterParameterHook trait.

Book documentation is updated to reflect this change and correct some other errors.

Added the book Python listings to the CI for testing.


pywr-core/src/parameters/py.rs
```diff
@@ -511,17 +599,32 @@ impl ParameterBuilder<f64> for PyClassParameterBuilder {
...
     ) -> Result<MaybeBuiltParameter<f64>, ParameterBuildError> {
-        let (has_before, has_after) = Python::attach(|py| {
+        let (has_before, after_type) = Python::attach(|py| {
             let has_before = self.class.getattr(py, "before").is_ok();
             let has_after = self.class.getattr(py, "after").is_ok();
-            (has_before, has_after)
-        });
+            let has_after_hook = self.class.getattr(py, "after_hook").is_ok();
+
+            let after_type = match (has_after, has_after_hook) {
```
Batch21: do impls for `u64` and `MultiValue` also need to do this?

pywr-core/src/parameters/py.rs
```diff
@@ -511,17 +599,32 @@ impl ParameterBuilder<f64> for PyClassParameterBuilder {
...
-            (true, true) => MetricConsumerPhase::Both,
-            (true, false) => MetricConsumerPhase::Before,
-            (false, true) => MetricConsumerPhase::After,
-            (false, false) => {
+            Ok((has_before, after_type))
+        })?;
+
+        let phase = match (has_before, after_type) {
+            (true, Some(AfterMethodType::After)) => MetricConsumerPhase::Both,
+            (true, Some(AfterMethodType::AfterHook)) => MetricConsumerPhase::Before,
```
Batch21: should this be `MetricConsumerPhase::Both` like the variant above?

pywr-core/src/parameters/py.rs
```diff
@@ -511,17 +599,32 @@ impl ParameterBuilder<f64> for PyClassParameterBuilder {
...
+            (true, None) => MetricConsumerPhase::Before,
+            (false, Some(AfterMethodType::After)) => MetricConsumerPhase::After,
+            (false, None) | (false, Some(AfterMethodType::AfterHook)) => {
                 return Err(ParameterBuildError::NoCalculationPhase {
                     detail: "PyClassParameterBuilder must have at least one of `before` or `after` methods defined."
```
Batch21: for the second case, should error say that an after hook requires a before?

Batch21 (changes requested):
Looks good - just a few small queries on this one

Batch21 (approved):

PR #809, opened 2026-09-16
feat: Allow referencing node-local parameters from global namespace.

jetuk:
Make a separate LocalParameterReference type that includes an optional node name. If this node name is provided it is used to find that parameter in that node's local parameter name space. I.e. it turns makes a reference with the given node name as the "parent".

If no "node" is provided the behaviour is as before.


pywr-schema/tests/local-parameter4.json
```diff
@@ -0,0 +1,303 @@
...
+      },
+      {
+        "from_node": "storage2",
+        "to_node": "demand2"
+      }
+    ],
+    "parameters": [
+      {
+        "meta": {
+          "name": "drought level"
```
Batch21, line 249: This matches the name of one of the local parameters. Should they have different names?

pywr-schema/src/metric.rs
```diff
@@ -648,6 +639,154 @@ impl ParameterReferenceBuilder {
...
+                }
+            }
+            None => UnresolvedMetricU64::ParameterValue { name, return_value },
+        };
+        Ok(m)
+    }
+    pub fn parameter_type(&self, args: &LoadArgs) -> Result<ParameterType, SchemaError> {
+        let parameter =
+            args.schema
+                .get_parameter_by_name(&self.name)
```
Batch21, line 725: This searches the global parameters. Shouldn't it get the node first and then query its local parameters?

pywr-schema/src/metric.rs
```diff
@@ -648,6 +639,154 @@ impl ParameterReferenceBuilder {
...
+    fn attribute(&self) -> String {
+        match &self.key {
+            Some(key) => key.clone(),
+            None => self.return_value.unwrap_or_default().to_string().to_lowercase(),
+        }
+    }
+}
+
+/// A builder for creating a [`ParameterReference`].
+pub struct LocalParameterReferenceBuilder {
```
Batch21, line 743: This isn't used yet. Is the plan to add a `all_local_parameters` flad to metric sets?

Batch21 (changes requested):
I think updates are also required in a couple of other places:
- `SchemaError::LocalParameterReferenceRequiresParent` needs updating
- the reference visitor needs to account for the option node name. I think the contents of `Reference::LocalParameter` and `ReferenceMut::LocalParameter` in visit.rs needs to become `{ node: Option<&'a mut String>, name: &'a mut String }`

pywr-schema/tests/local-parameter4.json
jetuk: this was intentional to check the namespaces work correctly.

pywr-schema/src/metric.rs
jetuk: Yes, I think so.

pywr-schema/src/metric.rs
```diff
@@ -648,6 +639,162 @@ impl ParameterReferenceBuilder {
...
+                }
+            }
+            None => UnresolvedMetricU64::ParameterValue { name, return_value },
+        };
+        Ok(m)
+    }
+    pub fn parameter_type(&self, args: &LoadArgs, parent: Option<&str>) -> Result<ParameterType, SchemaError> {
+        let parent = self.parent(parent)?;
+
+        let node = args
```
Batch21: I think this needs to account for reference to local params defined on virtual nodes. Something like:

```rust
let parameter = match args.schema.get_node_by_name(parent) {
    Some(node) => node.get_local_parameter(&self.name),
    None => args
        .schema
        .get_virtual_node_by_name(parent)
        .ok_or_else(|| SchemaError::NodeNotFound { name: parent.to_string() })?
        .get_local_parameter(&self.name),
}
.ok_or_else(|| SchemaError::ParameterNotFound { name: self.name.clone(), key: self.key.clone() })?;
```

pywr-schema/src/metric.rs
```diff
@@ -648,6 +639,162 @@ impl ParameterReferenceBuilder {
...
+
+    fn attribute(&self) -> String {
+        match &self.key {
+            Some(key) => key.clone(),
+            None => self.return_value.unwrap_or_default().to_string().to_lowercase(),
+        }
+    }
+}
+
+/// A builder for creating a [`ParameterReference`].
```
Batch21: docstring needs updating

Batch21 (changes requested):

Batch21 (approved):

PR #810, opened 2026-09-18
perf: Inform Clp about what has changed in the LP between solves.

jetuk:
The main C API does not permit setting the "start finish" options in the call to `dual`. By using these flags along with the "whatschanged" flag on the model Clp should be better informed of what we have changed.


coin-or-sys/build.rs
```diff
@@ -223,6 +223,8 @@ fn compile_clp() {
     builder.file(format!("{CLP_OSI_SRC_PATH}/OsiClpSolverInterface.cpp",));
     builder.include(CLP_OSI_SRC_PATH);
 
+    builder.file("src/clp_extensions.cpp");
```
Batch21, line 226: so apparently adding this below this line will force a rebuild when the file is changed:

```rust
println!("cargo:rerun-if-changed=src/clp_extensions.cpp");
```

Batch21 (changes requested):
I think we probably want a couple of simple unit test that check that flags are set correctly and factors are not reused when they shouldn't be.

jetuk:
I would like to rebase this on #825 to test the difference in performance.

Batch21 (approved):

PR #823, opened 2026-09-19
fix: Correct some TimeSeries references and names.

jetuk:


Batch21 (changes requested):
Only remaining change I can see is the filename of `pywr-book\src\external-data\timeseries.md`

Batch21 (approved):

PR #826, opened 2026-09-20
feat: Add ArrowStreamOutput.

jetuk:
A new output recorder that streams record batches in Arrow streaming format.


pywr-core/src/recorders/arrow_stream.rs
```diff
@@ -0,0 +1,586 @@
...
+        self.inner.flush()
+    }
+}
+
+fn make_schema(metric_set_name: &str, metrics: impl Iterator<Item = OutputMetric>) -> Schema {
+    let mut fields = vec![
+        Field::new("time_start", DataType::Timestamp(TimeUnit::Millisecond, None), false),
+        Field::new("time_end", DataType::Timestamp(TimeUnit::Millisecond, None), false),
+        Field::new("simulation_id", DataType::UInt64, false),
+        Field::new("label", DataType::Utf8, false),
```
Batch21, line 191: Do the field names here need to be reserved keys in case a metric name matches one of them?

Batch21 (approved):
one suggestion but otherwise LGTM

pywr-core/src/recorders/arrow_stream.rs
jetuk: We could just pre-populate `used_names` with those field names?

PR #827, opened 2026-09-21
feat(schema): Validate the scenario domain and references to its groups

Batch21:


pywr-schema/src/model.rs
```diff
@@ -358,6 +506,82 @@ pub struct ScenarioDomain {
...
+            .map(|(name, count)| ScenarioProblem::DuplicateGroupName {
+                name: name.to_string(),
+                count,
+            })
+            .collect();
+
+        problems.extend(self.groups.iter().flat_map(ScenarioGroup::problems));
+
+        if let Some(combinations) = &self.combinations {
+            problems.extend(self.combination_problems(combinations));
```
jetuk: This seems a bit odd to call the method with a reference to a field on self. Is there ever a need to validate the domain with another set of combinations?

jetuk (changes requested):
Just one question about signature of a method.

pywr-schema/src/model.rs
Batch21: no, I've changed this.

jetuk (approved):

PR #834, opened 2026-09-22
chore: Enforce cargo fmt.

jetuk:
- Fix all outstanding formatting issues.
- Add cargo fmt --check to GHA.


Batch21 (approved):
Only suggestions is that that fmt is run as the first job in the pipeline after checkout so that it reports error earlier

